PR #3, opened 2018-11-29
Coffee shop

eandreea:
Mersi de hint :), tocmai partea de fork nu am citit-o cand m-am uitat la partea de pull request.
Partea de app a ramas in starea de ieri, nu ruleaza conform cerintelor dupa cum v-am explicat in mail:  " am facut totul intr-un playground, dar surpriza a aparut cand am vrut sa rulez playgroundul la sfarsit cu tot cu partea de teste si playground support in el, cand am primit error: cannot load underlying module for 'XCTest' "
Aici trebuie sa vad ce solutii exista.


alexandru-calinoiu:
Please add minimum instructions on how to run the code, like the expected location for the file

CoffeeShop.playground/Contents.swift
```diff
@@ -1,6 +1,18 @@
 import Foundation
 import XCTest
 
+struct CoffeeShopApp {
+
+    init?(_ arguments: [String]) {
```
alexandru-calinoiu: Way to much code for the first test to pass.

You actually want to add as little code as possible to make the test pass, in this case parameter validation/parsing was not in the scope

CoffeeShop.playground/Contents.swift
```diff
@@ -30,3 +30,28 @@ class CoffeeShopAppTest: XCTestCase {
...
+
+class LocationTest: XCTestCase {
+
+    override func setUp() {
+    }
+
+    override func tearDown() {
+    }
+
+    func test_defaultLocation() {
```
alexandru-calinoiu: One should only write one test that fails.

CoffeeShop.playground/Contents.swift
pirvudoru: Why use `init?`?

CoffeeShop.playground/Contents.swift
```diff
@@ -31,6 +31,11 @@ class CoffeeShopAppTest: XCTestCase {
 
 CoffeeShopAppTest.defaultTestSuite.run()
 
+struct Location {
+    var x = 0.0
```
alexandru-calinoiu: Why do we want to have the location mutable? Is generally not a good idea to have this value objects muttable

CoffeeShop.playground/Contents.swift
```diff
@@ -1,65 +1,91 @@
...
+    }
+}
+
+LocationTest.defaultTestSuite.run()
+
+
+// MARK: - CoffeeShopApp
 struct CoffeeShopApp {
 
+    var user = Location()
```
pirvudoru: Why the variable name is user but the type is Location?

CoffeeShop.playground/Contents.swift
```diff
@@ -42,14 +42,26 @@ struct CoffeeShopApp {
     init?(_ arguments: [String]) {
         guard arguments.count == 3,
             let y = Double(arguments[0]),
-            let x = Double(arguments[1])
+            let x = Double(arguments[1]),
+            CoffeeShopApp.isFilename(arguments[2])
```
pirvudoru: This is not tested. Saw tests only for `isFilename`.
What is this call used for ?

CoffeeShop.playground/Contents.swift
```diff
@@ -58,6 +58,39 @@ class CoffeeShopTest: XCTestCase {
         XCTAssertEqual(coffeeShop.location.x, 47.5809)
         XCTAssertEqual(coffeeShop.location.y, -122.3160)
     }
+
+    func test_coffeeShopFromString() {
+        let coffeeShop = CoffeeShop("Starbucks Seattle,47.5809,-122.3160", separator: ",")
```
pirvudoru: Would prefer to have the constructor as dumb as possible. 

https://tamasgyorfi.net/2016/04/16/work-in-the-constructor-anti-pattern/

CoffeeShop.playground/Contents.swift
```diff
@@ -34,7 +34,7 @@ LocationTest.defaultTestSuite.run()
 
 
 // MARK: - CoffeeShop
-struct CoffeeShop {
+struct CoffeeShop: Equatable {
```
pirvudoru: Why is this needed?

CoffeeShop.playground/Contents.swift
```diff
@@ -166,6 +166,30 @@ class CoffeeShopTest: XCTestCase {
...
+                        Starbucks Moscow,55.752047,37.595242
+                        Starbucks Seattle2,47.5869,-122.3368
+                        Starbucks Rio De Janeiro,-22.923489,-43.234418
+                        Starbucks Sydney,-33.871843,151.206767
+                        """
+        let coffeeShops = CoffeeShop.coffeeShops(fromCSV: fileContent)
+        XCTAssertNotNil(coffeeShops)
+        XCTAssertEqual(coffeeShops!.count, 6)
+        let location = Location(x: -122.4, y: 47.6)
+        let orderedCoffeeShops = CoffeeShop.sort(coffeeShops: coffeeShops, byDistanceTo: location, ascending: true)
```
pirvudoru: This looks more like a procedural way of doing things, in which data is not grouped together with the logic that can act on it.

CoffeeShop.playground/Contents.swift
```diff
@@ -69,6 +78,17 @@ struct CoffeeShop: Equatable {
         return coffeeShops
     }
 
+    static func sort(coffeeShops: [CoffeeShop], byDistanceTo location: Location, ascending: Bool = true) -> [CoffeeShopWithDistance] {
+        var coffeeShopWithDistance = coffeeShops.map { CoffeeShopWithDistance(coffeeShop: $0, distance: $0.location.distanceTo(location)) }
+        coffeeShopWithDistance.sort {
+            if ascending {
+                return $0.distance < $1.distance
```
pirvudoru: What is the benefit of using $0 / $1 over named parameters ?

CoffeeShop.playground/Contents.swift
```diff
@@ -26,6 +26,9 @@ class CoffeeShopAppTest: XCTestCase {
     func test_applicationValidArguments() {
         let app = CoffeeShopApp(["47.6", "-122.4", "coffee_shop.csv"])
         XCTAssertNotNil(app)
+        XCTAssertEqual(app!.user.y, 47.6)
+        XCTAssertEqual(app!.user.x, -122.4)
+        XCTAssertEqual(app!.shopDataFilename, "coffee_shop.csv")
```
alexandru-calinoiu: At this point this test is not failing, it was written a little to late.

CoffeeShop.playground/Contents.swift
```diff
@@ -83,6 +83,24 @@ class CoffeeShopAppTest: XCTestCase {
         XCTAssertNil(app)
     }
 
+    func test_isFilename() {
+        let isFilename = CoffeeShopApp.isFilename("coffee_shop.csv")
+        XCTAssertTrue(isFilename)
+    }
+
+    func test_isNotFilename() {
+        let isFilename = CoffeeShopApp.isFilename("coffee_shop_csv")
```
alexandru-calinoiu: I will argue that this is not needed, `coffee_shop_csv` is a perfectly valid file name, it does not need an extension.
I think that only thing needed here was that the filename was not empty.

CoffeeShop.playground/Contents.swift
```diff
@@ -32,6 +32,23 @@ class LocationTest: XCTestCase {
...
+    override func setUp() {
+    }
+
+    override func tearDown() {
+    }
+
+    func test_coffeeShop() {
+        let coffeeShop = CoffeeShop(name: "Starbucks Seattle", location: Location(x: 47.5809, y: -122.3160))
+        XCTAssertEqual(coffeeShop.name, "Starbucks Seattle")
+        XCTAssertEqual(coffeeShop.location.x, 47.5809)
```
alexandru-calinoiu: You retesting the `Location` constructor, you've already done this.

CoffeeShop.playground/Contents.swift
```diff
@@ -32,6 +32,18 @@ class LocationTest: XCTestCase {
 
 LocationTest.defaultTestSuite.run()
 
+
+// MARK: - CoffeeShop
+struct CoffeeShop {
+    var name: String
```
alexandru-calinoiu: Again I think this structure should not be muttable.

CoffeeShop.playground/Contents.swift
```diff
@@ -166,6 +166,30 @@ class CoffeeShopTest: XCTestCase {
...
+        let location = Location(x: -122.4, y: 47.6)
+        let orderedCoffeeShops = CoffeeShop.sort(coffeeShops: coffeeShops, byDistanceTo: location, ascending: true)
+        var ordered = true
+        for i in 0..<coffeeShops!.count-1 {
+            let distance1 = orderedCoffeeShops[i].distance
+            let distance2 = orderedCoffeeShops[i+1].distance
+            if distance1 > distance2 {
+                ordered = false
+            }
+        }
```
pirvudoru: Is there any other way of comparing arrays in without doing this custom logic? One would not trust that the logic is accurate by glancing over tests.

CoffeeShop.playground/Contents.swift
```diff
@@ -210,6 +214,31 @@ class CoffeeShopTest: XCTestCase {
...
+        XCTAssertEqual(coffeeShops!.count, 6)
+        let location = Location(x: -122.4, y: 47.6)
+        let orderedCoffeeShops = CoffeeShop.sort(coffeeShops: coffeeShops!, byDistanceTo: location, ascending: false)
+        var ordered = true
+        for i in 0..<coffeeShops!.count-1 {
+            let distance1 = orderedCoffeeShops[i].distance
+            let distance2 = orderedCoffeeShops[i+1].distance
+            if distance1 < distance2 {
+                ordered = false
+            }
```
pirvudoru: This is duplicate as in test_coffeeShopsOrderedByDistanceToLocation

CoffeeShop.playground/Contents.swift
```diff
@@ -58,6 +58,39 @@ class CoffeeShopTest: XCTestCase {
         XCTAssertEqual(coffeeShop.location.x, 47.5809)
         XCTAssertEqual(coffeeShop.location.y, -122.3160)
     }
+
+    func test_coffeeShopFromString() {
```
alexandru-calinoiu: Same comment as before, way to many tests, the cycle should be way smaller, have a fail make it pass quickly

CoffeeShop.playground/Contents.swift
pirvudoru: Passing in booleans: https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

CoffeeShop.playground/Contents.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 import XCTest
 
 // MARK: - Location
-struct Location {
+struct Location: Equatable {
```
alexandru-calinoiu: You only make this structs Equatable because of the tests, this is usually a best practice, the production code should have meaning for production and should be driven by business requirements not implementation details.

CoffeeShop.playground/Contents.swift
```diff
@@ -315,17 +328,65 @@ class CoffeeShopAppTest: XCTestCase {
...
+                        Starbucks Seattle,47.5809,-122.3160
+                        Starbucks SF,37.5209,-122.3340
+                        Starbucks Moscow,55.752047,37.595242
+                        Starbucks Seattle2,47.5869,-122.3368
+                        Starbucks Rio De Janeiro,-22.923489,-43.234418
+                        Starbucks Sydney,-33.871843,151.206767
+                        """
+
+        guard let coffeeShopsWithDistance = app!.closestCoffeeShops(count: 3, coffeeShopsCSV: fileContent) else {
+            NSLog("Invalid data format")
```
pirvudoru: What is this used for?

CoffeeShop.playground/Contents.swift
```diff
@@ -315,17 +328,65 @@ class CoffeeShopAppTest: XCTestCase {
...
 
-    func test_test_isNotFilenameNoExtension() {
+    func test_isNotFilenameNoExtension() {
         let isFilename = CoffeeShopApp.isFilename("coffee_shop.")
         XCTAssertFalse(isFilename)
     }
+
+    func test_3ClosestCoffeeShops() {
+        let app = CoffeeShopApp(["47.6", "-122.4", "coffee_shop.csv"])
+        XCTAssertNotNil(app)
```
pirvudoru: Ref: http://defragdev.com/blog/?p=783

CoffeeShop.playground/Contents.swift
```diff
@@ -315,17 +328,65 @@ class CoffeeShopAppTest: XCTestCase {
...
+
+    for coffeeShopWithDistance in coffeeShopsWithDistance {
+        let distance = String(format: "%.4f", coffeeShopWithDistance.distance)
+        NSLog("\(coffeeShopWithDistance.coffeeShop.name),\(distance)")
+    }
+} else {
+    NSLog("invalid paramenters for Coffee Shop application")
+}
 
+print("end")
```
pirvudoru: What is this used for ?

CoffeeShop.playground/Contents.swift
```diff
@@ -166,6 +166,30 @@ class CoffeeShopTest: XCTestCase {
...
+                        Starbucks Rio De Janeiro,-22.923489,-43.234418
+                        Starbucks Sydney,-33.871843,151.206767
+                        """
+        let coffeeShops = CoffeeShop.coffeeShops(fromCSV: fileContent)
+        XCTAssertNotNil(coffeeShops)
+        XCTAssertEqual(coffeeShops!.count, 6)
+        let location = Location(x: -122.4, y: 47.6)
+        let orderedCoffeeShops = CoffeeShop.sort(coffeeShops: coffeeShops, byDistanceTo: location, ascending: true)
+        var ordered = true
+        for i in 0..<coffeeShops!.count-1 {
```
alexandru-calinoiu: This test is complicated.

It should be very easy to reason about a test, in this case I will simplify the preparation and only leave a couple of fake locations and make the assertion really easy and clear.

CoffeeShop.playground/Contents.swift
```diff
@@ -28,13 +28,22 @@ class LocationTest: XCTestCase {
         XCTAssertEqual(location.x, 1)
         XCTAssertEqual(location.y, 1)
     }
+
+    func test_distanceToLocation() {
```
alexandru-calinoiu: This is a good test, is small and clear. :+1:

CoffeeShop.playground/Contents.swift
```diff
@@ -28,13 +28,22 @@ class LocationTest: XCTestCase {
...
+    }
 }
 
 LocationTest.defaultTestSuite.run()
 
 
 // MARK: - CoffeeShop
 struct CoffeeShop: Equatable {
+
+    typealias CoffeeShopWithDistance = (coffeeShop: CoffeeShop, distance: Double)
```
alexandru-calinoiu: A tiny bit of functionality slipped in :)

CoffeeShop.playground/Contents.swift
```diff
@@ -315,17 +328,65 @@ class CoffeeShopAppTest: XCTestCase {
...
 
+var commandLineArguments = CommandLine.arguments
+commandLineArguments.removeFirst()
+
+done: if let app = CoffeeShopApp(commandLineArguments) {
+
+    let fileUrl = playgroundSharedDataDirectory.appendingPathComponent(app.shopDataFilename)
+    guard let fileContent = try? String(contentsOf: fileUrl) else {
+        NSLog("Unable to read \(app.shopDataFilename) from ~\\Documents\\Shared Playground Data")
+        break done
```
alexandru-calinoiu: I don't think this is required, I would expect to throw a custom exception if the input is invalid instead of using this goto like statement.

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
...
+            let x = Double(components[1]),
+            let y = Double(components[2])
+            else {
+                return nil
+        }
+
+        self.init(name: components[0], location: Location(x: x, y: y))
+    }
+
+    static func coffeeShops(fromCSV content: String) -> [CoffeeShop]? {
```
alexandru-calinoiu: This looks like a [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) can you refactor it to be more clear?

CoffeeShop.playground/Contents.swift
```diff
@@ -96,6 +63,125 @@ struct CoffeeShop: Equatable {
...
+    }
+
+} else {
+    NSLog("invalid paramenters for Coffee Shop application")
+}
+
+
+// MARK: - unit tests
+
+// uncomment to run tests
```
pirvudoru: Running tests should be straight forward and enforced or else they will be considered optional - tests are never optional (they prove that your code works).

Related to comments: 
https://refactoring.guru/smells/comments
https://blog.submain.com/field-guide-code-comments/

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
...
+        for line in lines {
+            guard let coffeeShop = CoffeeShop(line, separator: ",") else {
+                return nil
+            }
+            coffeeShops.append(coffeeShop)
+        }
+        return coffeeShops
+    }
+
+    static func sort(coffeeShops: [CoffeeShop], byDistanceTo location: Location, ascending: Bool = true) -> [CoffeeShopWithDistance] {
```
alexandru-calinoiu: I don't believe this static function should belong to this class, it has very little to do with the class that is hosted into

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
+import Foundation
+
+// MARK: - Location
+struct Location: Equatable {
```
pirvudoru: :+1:

alexandru-calinoiu (changes requested):
Overall:
- the use of static methods instead of having specialized services/classes makes code fell very procedural
- some unit tests try to do to much

Please review comments that have questions.

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
...
+    var x = 0.0
+    var y = 0.0
+
+    func distanceTo(_ location: Location) -> Double {
+        return sqrt( pow(self.x - location.x, 2) + pow(self.y - location.y, 2) )
+    }
+}
+
+
+typealias CoffeeShopWithDistance = (coffeeShop: CoffeeShop, distance: Double)
```
pirvudoru: How do you decide whether to use typealias or struct?

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
...
+
+    var name: String
+    var location: Location
+
+    init(name: String, location: Location) {
+        self.name = name
+        self.location = location
+    }
+
+    init?(_ line: String, separator: String) {
```
pirvudoru: What is the `_` used for here ?

CoffeeShop.playground/Contents.swift
```diff
@@ -0,0 +1,413 @@
...
+            } else {
+                return $0.distance > $1.distance
+            }
+        }
+        return coffeeShopWithDistance
+    }
+}
+
+// MARK: - CoffeeShopApp
+struct CoffeeShopApp {
```
pirvudoru: Prefer to have 1 class / file. It would be much easier to navigate.

pirvudoru (changes requested):
Hello. Thank you for submitting the solution. I'm not a swift developer. Please check the comments.

pirvudoru:
Also see: https://chris.beams.io/posts/git-commit/

CoffeeShop.playground/Contents.swift
eandreea: Ok, will refactor it. Thought that the arguments should be mandatory for the app and should be validated

CoffeeShop.playground/Contents.swift
eandreea: Yes, you are right

CoffeeShop.playground/Contents.swift
eandreea: yes....so i don't have to write the tuple description every time and use the typealias instead

CoffeeShop.playground/Contents.swift
eandreea: By "Malformed entries should cause the program to exit appropriately" request I thought that the app should not throw and error, that's why I logged the messages in the console. I'll refactor these parts

CoffeeShop.playground/Contents.swift
eandreea: Yeah....not really what i wanted in the first place, but forgot to refactor it after correctly reading the file.

CoffeeShop.playground/Contents.swift
eandreea: This should have been refactored with the other static fct

CoffeeShop.playground/Contents.swift
eandreea: The coffee shop app object should not initialise if the provided command line arguments are not as required (x, y, file_name). To run a coffee shop app a user location and coffee shop data file are required.

CoffeeShop.playground/Contents.swift
eandreea: Should have named it userLocation

CoffeeShop.playground/Contents.swift
eandreea: test_applicationNotEnoughArguments, test_applicationYNotDouble and test_applicationXNotDouble test the failable initializer

CoffeeShop.playground/Contents.swift
eandreea: I've made it as a convenience initializer for when we read the csv file and basically each line represents a coffee shop.

CoffeeShop.playground/Contents.swift
eandreea: In order to compare 2 instances of coffee shop, the class needs to conform to the Equatable protocol.

CoffeeShop.playground/Contents.swift
eandreea: Ok, I've over complicated things

CoffeeShop.playground/Contents.swift
eandreea: True, haven't thought of it this way. Should have compared the location property directly

CoffeeShop.playground/Contents.swift
eandreea: Yes, I've added the conformance to the Equatable protocol only for testing purposes. I'l keep this in mind when refactoring

CoffeeShop.playground/Contents.swift
eandreea: This is what I had in mind when i used the convenience initializer this way: "Create convenience initializers whenever a shortcut to a common initialization pattern will save time or make initialization of the class clearer in intent."
https://docs.swift.org/swift-book/LanguageGuide/Initialization.html

CoffeeShop.playground/Contents.swift
eandreea: Yes

CoffeeShop.playground/Contents.swift
eandreea: I've used swift shorthand argument names https://docs.swift.org/swift-book/LanguageGuide/Closures.html#ID97

CoffeeShop.playground/Contents.swift
eandreea: I should have used a more specific test here

CoffeeShop.playground/Contents.swift
eandreea: Returns the closest 3 coffee shops from the coffee shops cvs file content string

CoffeeShop.playground/Contents.swift
eandreea: I've commented them because I couldn't run the playground from the terminal. But since I found out yesterday that this was not required, I'll do some refactoring

CoffeeShop.playground/Contents.swift
eandreea: The struct is a coffee shop while the typealias is a renamed tuple which contains a coffee shop and a distance to the provided user location

CoffeeShop.playground/Contents.swift
eandreea: It means that the label of the argument is missing 
https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID166

CoffeeShop.playground/Contents.swift
eandreea: I'll refactor this

CoffeeShop.playground/Sources/CoffeeShopAppTest.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import XCTest
+
+public class CoffeeShopAppTest: XCTestCase {
+
+    public override func setUp() {
```
MihaiPantiru: @eandreea You should avoid overriding methods with no implementation. Is just "noise" to code base, no value added.

CoffeeShop.playground/Sources/CoffeeShop.swift
```diff
@@ -0,0 +1,33 @@
...
+    /// <Name>separator<Y Coordinate>separator<X Coordinate>
+    ///
+    /// - Parameter line: The string from where to extract data for the new instance.
+    /// - Parameter separator: The sequence that separates the data.
+    public init?(from line: String, separator: String) {
+        let compoments = line.components(separatedBy: separator)
+        guard !separator.isEmpty,
+            compoments.count == 3,
+            !compoments[0].isEmpty,
+            let x = Double(compoments[2]),
```
MihaiPantiru, line 26: @eandreea The results are wrong when running the app. This is happening because on read x is the last argument and y is the second one. After comparing with user location you get the results by comparing the user x with local y actually. Please use here x as second arg and y as the third argument.

CoffeeShop.playground/Sources/LocationTest.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import XCTest
+
+extension Location: Equatable {
```
MihaiPantiru, line 4: @eandreea The extension is not used on this file.

CoffeeShop.playground/Sources/CoffeeShopTest.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+import XCTest
+
+extension CoffeeShop: Equatable {
```
MihaiPantiru: @eandreea As a good practice, is not necessary to extend existing functionality just to write tests. For example, instead of comparing coffeeShop!.location with  Location(x: 47.5809, y: -122.3160), you can compare  coffeeShop!.location.x with 47.5809

CoffeeShop.playground/Sources/CoffeeShopTest.swift
```diff
@@ -0,0 +1,61 @@
...
+        let coffeeShop = CoffeeShop(name: "Starbucks Seattle", location: Location(x: 47.5809, y: -122.3160))
+        XCTAssertEqual(coffeeShop.name, "Starbucks Seattle")
+        XCTAssertEqual(coffeeShop.location, Location(x: 47.5809, y: -122.3160))
+    }
+
+    public func test_coffeeShopFromString() {
+        let coffeeShop = CoffeeShop(from: "Starbucks Seattle,-122.3160,47.5809", separator: ",")
+        XCTAssertNotNil(coffeeShop)
+        XCTAssertEqual(coffeeShop!.name, "Starbucks Seattle")
+        XCTAssertEqual(coffeeShop!.location, Location(x: 47.5809, y: -122.3160))
```
MihaiPantiru, line 17: @eandreea You should avoid using force unwrappers. I can very easily make this code crash. Is better to have a guard and fail earlier. Ex
guard let sut = coffeeShop else {
 // fail
}
XCTAssertEqual(sut.name, "Starbucks Seattle")

MihaiPantiru (changes requested):
@eandreea Very good progress, the code is easier to read and better structured.
I still have some remarks, please read my comments.

CoffeeShop.playground/Sources/CoffeeShop.swift
eandreea: Yes, as you can see, I've done this only in the last commit when I've noticed that the results were different from the ones before the refactoring. But this was request in the readme file.

CoffeeShop.playground/Sources/LocationTest.swift
eandreea: You are right, I've haven't placed it correctly, haven't looked at this from this point of view

CoffeeShop.playground/Sources/CoffeeShopTest.swift
eandreea: I agree, but in a previous comment it was stated that this functionality was already tested in the Location test, so it's just a duplicate. But to test that the CoffeeShop instance was correctly created I have to test the location parameter too... In the meantime i've seen that, after refactoring, only Location needs to conform to Equatable, so I'll remove the one from CoffeeShop

CoffeeShop.playground/Sources/LocationTest.swift
eandreea: Since I'm using the conformance of Location to Equatable protocol in both CoffeeShopTest and CoffeeShopAppTest, and I have to put the code only in one place, I support my first decision to put it in LocationTest. If the equality test would have been required in only one location, then I would support that the code be moved in that file

CoffeeShop.playground/Sources/CoffeeShopTest.swift
eandreea: Yeah, big mistake on my part, this was due to my forgetting to put continueAfterFailure = false in the test

CoffeeShop.playground/Sources/CoffeeShopTest.swift
MihaiPantiru: Great. Sorry for contradictory comments, we didn't sync the comments across the team. As long this is the result of another comment than is ok for me.

CoffeeShop.playground/Sources/LocationTest.swift
MihaiPantiru: What about having the Equatable conformance in a separated file if is needed in multiple places?

eandreea: Of course that works too, but I don't consider it necessary here. If something more complex was required, then the separate file would be best.

CoffeeShop.playground/Contents.swift
alexandru-calinoiu: `exit appropriately` can mean that you throw a explicit error if the user input is not right.

The best explanation I've found about this pattern is in the rust documentation file:///home/calin/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/book/2018-edition/ch09-03-to-panic-or-not-to-panic.html, panic! is like throwing an exception.

alexandru-calinoiu:
Closing this as we will followup via email.

PR #4, opened 2018-12-30
added code for Coffee Addicts test

alexandravirvara:


src/model/Coordinate.kt
```diff
@@ -0,0 +1,65 @@
...
+     *
+     **** arg: the String to be converted to a Double
+     **** returns the Double corresponding to the given coordinate as a String, if the coordinate is valid,
+     *              or an invalid coordinate otherwise
+     */
+    private fun convertToDouble(arg: String): Double {
+        return try {
+            arg.toDouble()
+        } catch (nfe: NumberFormatException) {
+            maxLongitude + 1
```
alexandru-calinoiu, line 62: Not sure what this is used for?

src/model/FileParser.kt
```diff
@@ -0,0 +1,93 @@
+import java.io.File
+
+/**
+ * Defines a FileParser object, containing different operations on the content of a given .csv file
+ *
+ * created by AlexandraV on 12/25/2018
+ */
+data class FileParser(var fileName: String, var userCoordinate: Coordinate) {
```
alexandru-calinoiu, line 8: I don't believe this file is used anymore. You should remove it as it fails to compile now..

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+import Coordinate
+import java.io.File
+
+/**
+ * Defines a parser.FileParser object, containing different operations on the content of a given .csv file
+ *
+ * created by AlexandraV on 12/25/2018
+ */
+data class FileParser(var fileName: String, var userCoordinate: Coordinate) {
+    val coffeeShopsList: MutableList<CoffeeShop> = mutableListOf()
```
alexandru-calinoiu, line 13: Most probably this should be private also, any reason is not?

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * inserts a CoffeeShop object into the list. The first ${n} elements of the list must be the first ${n} closest shops to the user
+     * and they must be ordered by distance, from the closest to the furthest
+     *
+     * *** coffeeShopName: the name of the coffeeShop
+     * *** coffeeShopCoordinateX: the X coordinate of the coffeeShop
+     * *** coffeeShopCoordinateY: the Y coordinate of the coffeeShop
+     * *** n : the number of elements from the beginning of the list, ordered by their distance to the user
+     */
+    fun insertShopIntoList(coffeeShopName: String, coffeeShopCoordinateX: String, coffeeShopCoordinateY: String, n: Int) {
```
pirvudoru, line 67: Why should this method be public?

src/model/FileParser.kt
```diff
@@ -0,0 +1,93 @@
...
+     */
+    fun parseCoffeeShopsList(n: Int) {
+        val invalidCoffeeShopsList: MutableList<String> = mutableListOf()
+        if (lines != null) {
+            var lineNumber = 1
+            lines.forEach { it ->
+                val result: List<String> = it.split(",").map { it.trim() }
+                if (isCSVLineValid(result)) {
+                    val shopCoordinate = Coordinate(result[1], result[2])
+                    insertShopIntoList(CoffeeShop(result[0], shopCoordinate.distanceTo(userCoordinate.xCoordinate, userCoordinate.yCoordinate)), n)
```
pirvudoru, line 24: Why prefer passing in x and y to distanceTo instead of a coordinate object?

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+        if (n > coffeeShopsList.size && !coffeeShopInserted) {
+            coffeeShopsList.add(coffeeShop)
+        }
+    }
+
+    /**
+     * Checks if a line is valid, of the form: CoffeeShopName, X ShopCoordinate, Y ShopCoordinate
+     */
+    fun isCSVLineValid(line: List<String>): Boolean {
+        if (line.size != 3) {
```
alexandru-calinoiu, line 93: This code is very hard to followup, [multi returns](https://www.yegor256.com/2015/08/18/multiple-return-statements-in-oop.html) statements are considered a bad practice, can you think of a way to avoid them?

src/parser/FileParser.kt
alexandru-calinoiu, line 67: A couple of comments about this methods signature:

- it does not have an access modifier, most likely should be private
- it receives x, and y and strings although there is a model to hold Coordinates, why is that?
- the `n` parameter not very expressive, I think it means the number of closes shops, the naming can be improved.

test/model/CoordinateTest.kt
```diff
@@ -0,0 +1,42 @@
...
+fun isCoordinateValid_Valid() {
+    val coordinate1 = Coordinate("45.0", "30.02")
+    val coordinate2 = Coordinate("0.0", "0.3")
+    val coordinate3 = Coordinate("-90.0", "180.0")
+    val coordinate4 = Coordinate("90.0", "-180.0")
+
+    assert(coordinate1.isCoordinateValid())
+    assert(coordinate2.isCoordinateValid())
+    assert(coordinate3.isCoordinateValid())
+    assert(coordinate4.isCoordinateValid())
```
pirvudoru, line 35: Sometimes it's ok to have multiple asserts per test, but in this case, the asserts/acts are independent so it's preferable to separate them out.

test/model/CoordinateTest.kt
```diff
@@ -0,0 +1,42 @@
+import org.testng.annotations.Test
+import kotlin.test.assertEquals
+
+/**
+ * created by AlexandraV on 12/25/2018
+ */
+@Test
+fun isCoordinateValid_bothInvalid() {
+    val coordinate = Coordinate("877.0", "956.256")
+    assert(!coordinate.isCoordinateValid()) { "x in [-90, 90], y in [-180, 180]" }
```
pirvudoru, line 10: Why would the constructor allow you to create the object if it the values are invalid?

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * *** coffeeShopName: the name of the coffeeShop
+     * *** coffeeShopCoordinateX: the X coordinate of the coffeeShop
+     * *** coffeeShopCoordinateY: the Y coordinate of the coffeeShop
+     * *** n : the number of elements from the beginning of the list, ordered by their distance to the user
+     */
+    fun insertShopIntoList(coffeeShopName: String, coffeeShopCoordinateX: String, coffeeShopCoordinateY: String, n: Int) {
+        val coffeeShop = CoffeeShop(coffeeShopName, Coordinate(coffeeShopCoordinateX, coffeeShopCoordinateY), Coordinate(userCoordinate.x, userCoordinate.y))
+        var coffeeShopInserted = false
+        // check if the current coffee shop enters the "top n" coffee shops according to their distance and insert it on the right position
```
alexandru-calinoiu, line 70: Code comments are considered a [code smell](https://blog.codinghorror.com/code-smells/). The best practice is to focus on the `why` not on the `what`, but better yet the code should be refactored to not need comments.

Do you really use this amount of comments in your production code?

test/model/CoordinateTest.kt
```diff
@@ -0,0 +1,42 @@
+import org.testng.annotations.Test
+import kotlin.test.assertEquals
+
+/**
+ * created by AlexandraV on 12/25/2018
+ */
+@Test
+fun isCoordinateValid_bothInvalid() {
+    val coordinate = Coordinate("877.0", "956.256")
```
pirvudoru, line 9: What are the pros/cons of having `string` vs `double` passed into the constructor of "Coordinate" ?

src/model/CoffeeShop.kt
```diff
@@ -0,0 +1,19 @@
+/**
+ * Defines a CoffeeShop object, containing a CoffeeShop name, its coordinates and a given user's coordinates
+ *
+ * created by AlexandraV on 12/26/2018
+ */
+data class CoffeeShop(val shopName: String, val shopCoordinate: Coordinate, val userCoordinate: Coordinate) : Comparable<CoffeeShop> {
```
pirvudoru, line 6: What if you want to compare the same coffee shop to more than one user? How would the CoffeeShop class look like then?

src/CoffeeAddictsMain.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+fun main(args: Array<String>) {
+    if (!areCommandLineArgumentsValid(args)) {
+        return
+    } else {
+        // read the coordinate of the user
+        val userCoordinate = Coordinate(args[0], args[1])
+
+        // read content of .csv file
+        val fileParser = FileParser(args[2], userCoordinate)
```
pirvudoru, line 15: What is the relation between a "FileParser" and a `userCoordinate`?

src/CoffeeAddictsMain.kt
```diff
@@ -0,0 +1,44 @@
...
+        fileParser.parseCoffeeShopsList(3)
```
pirvudoru, line 16: By the looks of the usage of this method, it as side effects and by passing in `3` as argument looks like does more things that it should. One would expect that `parseCoffeeShops` would process some kind of serialized data and return a list of coffee shops objects.

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                if (coffeeShopsList.size > 0) {
+                    coffeeShopsList.forEach { println("${it.shopName}, ${it.shopDistanceToUser}") }
+                    println("\nThe number of CoffeeShops you requested ($n) exceeds the number of valid shops in the .csv file. Only ${coffeeShopsList.size} result(s) printed")
+                } else {
+                    println("There are no valid coffee shops in $fileName to be displayed")
+                }
+            }
+        }
+        // write all invalid lines (if there are any) in a file
```
pirvudoru, line 47: Re comments:

https://refactoring.guru/smells/comments
https://blog.submain.com/field-guide-code-comments/

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+        if (!invalidCoffeeShopsList.isEmpty()) {
+            println("\nAll invalid lines found in $fileName are written in invalid_coffee_shops.txt")
+            File("invalid_coffee_shops.txt").bufferedWriter().use { out ->
+                invalidCoffeeShopsList.forEach {
+                    out.write(it)
+                }
+            }
+        }
```
pirvudoru, line 55: This method is quite hard to follow and does more things that it should. One shouldn't need comments to explain what the code does.

src/model/FileParser.kt
```diff
@@ -0,0 +1,93 @@
...
+                } else {
+                    invalidCoffeeShopsList.add("Line $lineNumber: $it \n")
+                }
+                lineNumber++
+            }
+
+            if (n <= coffeeShopsList.size) {
+                coffeeShopsList.subList(0, n).forEach {
+                    println("${it.shopName}, ${it.shopDistanceToUser}")
```
pirvudoru, line 33: What is the problem you are trying to solve by having `println`s all over the place?

pirvudoru (changes requested):
Thanks for submitting your solution. 

Please check comments and the following links:

What is TDD: http://agiledata.org/essays/tdd.html
How to write great Git commits: https://chris.beams.io/posts/git-commit/

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+     * *** coffeeShopCoordinateX: the X coordinate of the coffeeShop
+     * *** coffeeShopCoordinateY: the Y coordinate of the coffeeShop
+     * *** n : the number of elements from the beginning of the list, ordered by their distance to the user
+     */
+    fun insertShopIntoList(coffeeShopName: String, coffeeShopCoordinateX: String, coffeeShopCoordinateY: String, n: Int) {
+        val coffeeShop = CoffeeShop(coffeeShopName, Coordinate(coffeeShopCoordinateX, coffeeShopCoordinateY), Coordinate(userCoordinate.x, userCoordinate.y))
+        var coffeeShopInserted = false
+        // check if the current coffee shop enters the "top n" coffee shops according to their distance and insert it on the right position
+        for (i in 0..(minOf(n, coffeeShopsList.size) - 1)) {
+            if (coffeeShop < coffeeShopsList[i]) {
```
alexandru-calinoiu, line 72: This is a interesting approach.

Unfortunately for me it does not make any sens to compare coffee shops, because actually they get compared based on the distance from the current user.

There a couple of things that I think can be improved:
- coffe shop object should not hold the coordonates of the user, it's not its concern
- coffe shop's can be comparable but not to get the distance, I can imagine a dsl like `userCoordonates.distanceTo(coffeShoopCoordonates)`. The distance calculation should be the concern of some other class.

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val lines: List<String>? = File(fileName).useLines { it.toList() } // the reader is closed automatically
+
+    /**
+     *  parses the lines of the .csv file and converts them into CoffeeShops objects, if possible
+     *  All the lines that cannot be converted to CoffeeShops objects will be written into  invalid_coffee_shops.txt file
+     */
+    fun parseCoffeeShopsList(n: Int) {
+        val invalidCoffeeShopsList: MutableList<String> = mutableListOf()
+        if (lines != null) {
+            var lineNumber = 1
```
alexandru-calinoiu, line 23: No point on holding the `lineNumber` one can use [forEachIndexed()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each-indexed.html)

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+                val result: List<String> = it.split(",").map { it.trim() }
+                if (isCSVLineValid(result)) {
+                    insertShopIntoList(result[0], result[1], result[2], n)
+                } else {
+                    if (!it.trim().isEmpty()) invalidCoffeeShopsList.add("Line $lineNumber: $it \n")
+                }
+                lineNumber++
+            }
+
+            if (n <= coffeeShopsList.size) {
```
alexandru-calinoiu, line 34: I don't think the list size can be larger that n, as the insertShopIntoList has `minOf(n, coffeeShopsList.size) - 1`

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+import java.io.File
+
+/**
+ * Defines a parser.FileParser object, containing different operations on the content of a given .csv file
+ *
+ * created by AlexandraV on 12/25/2018
+ */
+data class FileParser(var fileName: String, var userCoordinate: Coordinate) {
+    val coffeeShopsList: MutableList<CoffeeShop> = mutableListOf()
+    private val lines: List<String>? = File(fileName).useLines { it.toList() } // the reader is closed automatically
```
alexandru-calinoiu, line 14: What about using a delegate here like [lazy](https://kotlinlang.org/docs/reference/delegated-properties.html#lazy) do you think that will be appropriate?

src/parser/FileParser.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     *  parses the lines of the .csv file and converts them into CoffeeShops objects, if possible
+     *  All the lines that cannot be converted to CoffeeShops objects will be written into  invalid_coffee_shops.txt file
+     */
+    fun parseCoffeeShopsList(n: Int) {
```
alexandru-calinoiu, line 20: This method dose so much more that the name suggests:
- it will parse file
- validate file
- calculate results
- print the results and errors

Let's talk about how to better express this flow in code, do you have some thoughts?

alexandru-calinoiu (commented):
Good effort, I like the fact that the code is broke down into multiple classes and it tries to separate concerns.

The big miss is that the test driven requirement from the email was totally ignored, there is no evidence that the red-green-refactor cycle was used.

alexandru-calinoiu:
Followeup via email.

PR #6, opened 2019-02-07
Coffee Shop 

elimelec:


app/tests/LocationTest.kt
```diff
@@ -0,0 +1,11 @@
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class LocationTest {
+    @Test
+    fun `test location`() {
+        val location = Location(12.51, -50.12)
+        assertEquals(12.51, location.x)
```
alexandru-calinoiu, line 8: If `Location` is a data class, it has some useful properties that will allow you to write only one assert.

app/src/Location.kt
```diff
@@ -0,0 +1 @@
+data class Location(val x: Double, val y: Double)
```
alexandru-calinoiu: Not implying that you made a mistake (this applies to all my comments by the way) but why choose `Double` over a `Float`?

app/tests/CoffeeShopTest.kt
```diff
@@ -0,0 +1,23 @@
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class CoffeeShopTest {
+    @Test
+    fun `test coffee shop`() {
+        val shop = CoffeeShop("Starbucks", Location(21.51, -74.58))
+        assertEquals("Starbucks", shop.name)
+//        assertEquals(21.51, shop.x)
```
alexandru-calinoiu: This comments are not very usefull

app/tests/CoffeeShopTest.kt
```diff
@@ -0,0 +1,23 @@
...
+//        assertEquals(-74.58, shop.y)
+        assertEquals(21.51, shop.location.x)
+        assertEquals(-74.58, shop.location.y)
+    }
+
+    fun `test coffee shop without using location`() {
```
alexandru-calinoiu, line 14: Good point, but still over work, you could have simply choose one path and if the need had arrisen write this second constructor.

app/tests/CoffeeShopTest.kt
```diff
@@ -0,0 +1,23 @@
...
+        val shop = CoffeeShop("Starbucks", 21.51, -74.58)
+        assertEquals("Starbucks", shop.name)
+//        assertEquals(21.51, shop.x)
+//        assertEquals(-74.58, shop.y)
+        assertEquals(21.51, shop.location.x)
```
alexandru-calinoiu: Accessing `x` and `y` like this makes me thing of the [feature envy code smell](https://duckduckgo.com/?q=feature+envy&t=ffab&atb=v129-2&ia=web)

app/tests/LocationTest.kt
```diff
@@ -0,0 +1,11 @@
...
+        assertEquals(-50.12, location.y)
```
alexandru-calinoiu, line 9: I will argue that `x` and `y` are not great names for locations, why not go with `longitude` and `latitude` ?

app/tests/DistanceTest.kt
```diff
@@ -0,0 +1,20 @@
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class DistanceTest {
+    @Test
+    fun `test distance from examples`() {
```
alexandru-calinoiu: To much of a test, a single example would have sufficed. And then consider doing a test for some edge case if any

app/tests/DistanceTest.kt
```diff
@@ -2,19 +2,31 @@ import org.junit.jupiter.api.Test
 import kotlin.test.assertEquals
 
 class DistanceTest {
+
     @Test
-    fun `test distance from examples`() {
+    fun `test distance from examples (user - seattle2)`() {
```
alexandru-calinoiu, line 9: Like the fact that you have 1 assert per test :+1: 

Still don't think is a good idea to add this many tests at this point.

app/src/Location.kt
```diff
@@ -1 +1,10 @@
+import kotlin.math.abs
+
 data class Location(val x: Double, val y: Double)
+
+fun distance(location1: Location, location2: Location): Double {
+    var (x1, y1) = location1
```
alexandru-calinoiu: Destructuring the `Location` class very cool.

local `x`'s and `y`'s should be made immutable

app/src/CoffeeShop.kt
```diff
@@ -13,3 +13,8 @@ fun String.toCoffeeShop(): CoffeeShop {
         throw IllegalArgumentException("row is invalid csv: $this")
     }
 }
+
+fun Array<String>.toCoffeeShops(): Array<CoffeeShop> {
```
alexandru-calinoiu: Why choose an `Array` over a `List`?

app/src/CoffeeShop.kt
```diff
@@ -14,7 +14,7 @@ fun String.toCoffeeShop(): CoffeeShop {
     }
 }
 
-fun Array<String>.toCoffeeShops(): Array<CoffeeShop> {
-    return this.map { it.toCoffeeShop() }.toTypedArray()
+fun List<String>.toCoffeeShops(): List<CoffeeShop> {
```
alexandru-calinoiu: Ha :) Why choose a `List` over an `Array`?

app/tests/CoffeeShopWithDistanceTest.kt
```diff
@@ -0,0 +1,10 @@
+import org.junit.jupiter.api.Test
+
+class CoffeeShopWithDistanceTest {
+    @Test
+    fun `test coffee shop with distance instance`() {
```
alexandru-calinoiu, line 11: Not really sure what this test was for initially, maybe I will understand later

app/tests/LocationTest.kt
```diff
@@ -8,4 +8,11 @@ class LocationTest {
         assertEquals(12.51, location.x)
         assertEquals(-50.12, location.y)
     }
+
+    @Test
+    fun `test location equality`() {
+        val location1 = Location(12.51, -50.12)
+        val location2 = Location(12.51, -50.12)
+        assert(location1 == location2)
```
alexandru-calinoiu: You wrote a test for a kotlin feature, I think we can leave the kotlin developers write tests for their own features.

app/src/CoffeeShopWithDistance.kt
```diff
@@ -1,2 +1,5 @@
 class CoffeeShopWithDistance(val shop: CoffeeShop, val location: Location) {
-}
\ No newline at end of file
+    val distance: Double by lazy {
```
alexandru-calinoiu: A `distance` class has a `distance` property intern.

app/tests/CoffeeShopWithDistanceTest.kt
```diff
@@ -28,4 +28,11 @@ class CoffeeShopWithDistanceTest {
         val other = CoffeeShopWithDistance(shop, location)
         assert(distance == other)
     }
+
+    @Test
+    fun `test inequality`() {
+        val otherLocation = Location(20.12, 50.8)
+        val other = CoffeeShopWithDistance(shop, otherLocation)
+        assert(distance != other)
```
alexandru-calinoiu: I will argue that this test is somehow redundant.

app/tests/CoffeeShopWithDistanceTest.kt
```diff
@@ -36,6 +36,7 @@ class CoffeeShopWithDistanceTest {
         assert(distance != other)
     }
 
+    @Test
```
alexandru-calinoiu: Obviously was not run before :)

app/tests/CoffeeShopAppTest.kt
```diff
@@ -0,0 +1,11 @@
+import org.junit.jupiter.api.Test
+
+class CoffeeShopAppTest {
+    @Test
+    fun `test coffee shop app instance`() {
+        val app = CoffeShopApp(47.6, -122.4, "coffee_shop.csv")
+        app.x = 47.6
```
alexandru-calinoiu: Not sure why you made variables writable.

app/src/CoffeeShopApp.kt
```diff
@@ -1,6 +1,6 @@
 class CoffeeShopApp (val x: Double, val y: Double, val filename: String) {
     fun getNearestShops(): List<CoffeeShopWithDistance> {
-        return listOf<CoffeeShopWithDistance>()
+        return listOf()
```
alexandru-calinoiu: You can shorted it even more:

```
     fun getNearestShops(): List<CoffeeShopWithDistance> = listOf()
```
Or better yet try to use `emptyList()`

app/tests/CoffeeShopAppTest.kt
```diff
@@ -24,6 +24,8 @@ class CoffeeShopAppTest {
     fun `test console output with example`() {
         val app = CoffeeShopApp(47.6, -122.4, "coffee_shops.csv")
         val output  = app.getOutput()
-        assertEquals("Starbucks Seattle2,0.0645\nStarbucks Seattle,0.0861\nStarbucks SF,10.0793", output)
+
+        val ls = System.lineSeparator()
+        assertEquals("Starbucks Seattle2,0.0645${ls}Starbucks Seattle,0.0861${ls}Starbucks SF,10.0793", output)
```
alexandru-calinoiu, line 31: :+1: did not knew about it.

app/src/CoffeeShopApp.kt
```diff
@@ -0,0 +1,33 @@
+import java.io.File
+import java.io.FileNotFoundException
+
+class CoffeeShopApp (val x: Double, val y: Double, val filename: String) {
+    fun getNearestShops(): List<CoffeeShopWithDistance> {
```
alexandru-calinoiu: This does not make sens to be public, you should limit the public surface, read more about it `Indecent Exposure` code smell.

app/src/CoffeeShopApp.kt
```diff
@@ -0,0 +1,33 @@
...
+        catch (e: FileNotFoundException) {
+            return "$filename was not found."
+        }
+        catch (e: CouldNotParseCsvRowException) {
+            return "$filename is not a valid file."
+        }
+    }
+}
+
+fun readAllLinesAsCsvRows(filename: String): List<String> {
```
alexandru-calinoiu: Again this method should not be public.

alexandru-calinoiu (commented):
This is a decent implementation.

But the goal of the tdd was largely missed, you basically knew the implementation from the beginning and started with testing the details.

To get a better feel of how this can be properly done, you can check out this video https://www.youtube.com/watch?v=qkblc5WRn-U

Another nice touch would have been a description for the pr to  help the reviewer with setting up the environment.

Please review the comments and response when appropriate (they where done in the order of the commits and some of them migt make sens)

app/src/CoffeeShopApp.kt
```diff
@@ -0,0 +1,33 @@
...
+            }
+            else {
+                return shops.joinToString(separator = System.lineSeparator())
+            }
+
+        }
+        catch (e: FileNotFoundException) {
+            return "$filename was not found."
+        }
+        catch (e: CouldNotParseCsvRowException) {
```
alexandru-calinoiu, line 25: Can you think of some other ways to have resilience but have a nicer user experience?

app/src/Location.kt
elimelec: I use `Double` by default since it has more precision and it's the default literal interpretation in most languages and never think about it if the need doesn't arise.

app/src/CoffeeShop.kt
elimelec: `Array` happened to be the first class I thought about when I wrote the test. I just needed a list like data type.

app/src/CoffeeShop.kt
elimelec: When I actually wrote code I noticed that stream functions return `List`, there was no point to always convert the `List` to `Array`, or to use different types across the app when one can handle everything just fine.

app/tests/LocationTest.kt
elimelec: You're right. I'll remove them.

app/tests/CoffeeShopAppTest.kt
elimelec: It was mistake, I wanted to write `assert(47.6, app.x)`... 😬

app/src/CoffeeShopApp.kt
elimelec: Nice, didn't know this.

app/src/CoffeeShopApp.kt
elimelec: I forgot about this. Will update the code.

app/tests/LocationTest.kt
elimelec: Are you thinking about `@ParameterizedTest`?

app/src/CoffeeShopWithDistance.kt
elimelec: I don't understand what you want to say.

app/tests/LocationTest.kt
elimelec: I don't think `x` and `y` are bad given that the specification uses `x` and `y` and this notation is used often by programmers for various points. `longitude` and `latitude` could be a better choice for a location, but I'd keep `x/y` to be consistent with the docs.

elimelec: One exception could be the usage output: `Usage: <program name> <user x coordinate> <user y coordinate> <shop data filename>.`

I think it would be easier to understand this message with `longitude/latitude` if you are not a developer.

app/src/CoffeeShopApp.kt
elimelec: I could ignore the bad rows from the csv file. The user could still use the app with the other available coffee shops.

app/tests/CoffeeShopTest.kt
elimelec: Ok, I'll keep this in mind.

app/tests/CoffeeShopWithDistanceTest.kt
elimelec: I agree.

app/tests/CoffeeShopWithDistanceTest.kt
elimelec: I tried to write only enough test to fail.

app/tests/LocationTest.kt
alexandru-calinoiu: No, actually I was thinking about the free equality operator that you also found out about later on in the pr.

app/src/CoffeeShopApp.kt
alexandru-calinoiu: Yeah, good point, also should consider to report the bad rows (all of them) I always find it annoying when apps keep feeding me input errors one by one.

alexandru-calinoiu:
Thank you for your hard work, will followup shortly via email.

PR #7, opened 2019-02-13
Coffee shop kotlin eduard dudu

eduardDudu:


coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
```diff
@@ -0,0 +1,72 @@
...
+    @Throws(Exception::class)
+    fun parseCSV(input: String?): List<Shop>? {
+
+        val csvSplitter = ","
+        var array: List<Shop>?
+        array = ArrayList()
+
+        File(input).forEachLine {
+            if (it.contains(",")) {
+                val vars = it.split(csvSplitter.toRegex()).dropLastWhile({ it.isEmpty() }).toTypedArray()
```
alexandru-calinoiu, line 19: The naming can be improved here, reading `vars` is hard to figure out what it contains.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
```diff
@@ -0,0 +1,72 @@
...
+                if (isValidLine(vars)) {
+                    if (isValidName(vars)) {
+                        if (isValidDouble(vars[1]) && isValidDouble(vars[2])) {
+
```
alexandru-calinoiu, line 23: The style is not very consistent, this new lines here don't add any value

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
```diff
@@ -0,0 +1,72 @@
...
+                                array.add(getShop(vars))
```
alexandru-calinoiu, line 24: This if statements represent validation logic, is there a better way to express this?

On a more funnier note, this reminds me of https://twitter.com/tom_enebo/status/476736105148989440

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
```diff
@@ -0,0 +1,72 @@
...
+
+                        } else {
+                            Log.e("ReadingFile", "Line: $it. Latitude or Longitude invalid.")
+                        }
+                    } else {
+                        Log.e("ReadingFile", "Line: $it. Name of the location invalid.")
+                    }
+                } else {
+                    /* Invalid line, continue with next one*/
```
alexandru-calinoiu, line 33: This comment actually explains the logic this is a good thing, the really bad thing si that we need a comment here the code should be reorganized in such a way that this line should not be required.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
```diff
@@ -0,0 +1,72 @@
...
+        return array
+    }
+
+
+    fun isValidLine(vars: Array<String>): Boolean{
+        return (vars.size == 3)
+    }
+
+    fun isValidName(vars: Array<String>): Boolean{
+        return vars[0].isNotEmpty()
```
alexandru-calinoiu, line 50: Holding a csv line in a array has obvious drawbacks like the fact that you mentally have to know the index of each column, is there a better approach to this?

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/Objects/Shop.kt
```diff
@@ -0,0 +1,10 @@
+package ro.eduarddudu.coffeeshop.Objects
+
+import android.location.Location
+
+class Shop {
```
alexandru-calinoiu, line 5: Having a Shop declared this way means that it's properties are mutable, this is usually a bad idea, can you tell me why?

Also is there is a better alternative?

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/ShopFinder.kt
```diff
@@ -0,0 +1,61 @@
...
+object ShopFinder {
+
+    @Throws(Exception::class)
+    fun findClosestLocations(latitude: Double, longitude: Double, path: String): List<Shop> {
+
+        val shops = CSVReader.parseCSV(path)
+                ?: throw Exception("Input file can not be parsed.")
+
+        if (shops.size == 0) {
+            throw Exception("Invalid input file.")
```
alexandru-calinoiu, line 16: One cannot have an empty file ?

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/ShopFinder.kt
```diff
@@ -0,0 +1,61 @@
...
+
+            return getfirstMax3Elements(sortedshops)
+        }
+    }
+
+    fun calculateDistance(src: Location, dest: Location?): Float{
+        return src.distanceTo(dest)
+    }
+
+    fun getfirstMax3Elements(list: List<Shop>): ArrayList<Shop> {
```
alexandru-calinoiu, line 43: There are [easier ways](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/take.html) to do this is kotin

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/ShopFinder.kt
```diff
@@ -0,0 +1,61 @@
...
+                ?: throw Exception("Input file can not be parsed.")
+
+        if (shops.size == 0) {
+            throw Exception("Invalid input file.")
+        }
+
+        if (shops.size == 1) {
+            return shops
+        } else {
+            val userLocation = Location("userLocation")
```
alexandru-calinoiu, line 22: The `Location` class is in the android package, this means that you have a dependency on the android platform and your code won't run outside it. 

Usually is a good idea to keep your business logic and platform independent as possible, do you happen to know why?

alexandru-calinoiu (commented):
Thank you for your pull request.

Overall I would think this is a not a very object oriented approach to solving the problem is more procedural, that itself is not a bad thing but it raises some questions about the separation of concerns.

Please take a look at my questions and let me know what you think.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/ShopFinder.kt
eduardDudu: I know are another ways to calculate the distance between 2 cardinal points. But this was my choose in that moment. I understand is better to don`t have the dependency on a specific platform for the reason you can use in other languages projects.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/ShopFinder.kt
eduardDudu: I didn`t know about this. Thank you.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/Objects/Shop.kt
eduardDudu: Can be declared like immutable with final, but, i used property distance which is calculated after the object Shop is initialized. Only if the distance is calculated when the object is created then the calss can be immutable.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
eduardDudu: I can see the way with use the String line, but when a check is made need the array again. How can you suggest to use this ? Thank you

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
eduardDudu: I used more ifs to have a detalied error for each case.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
eduardDudu: Can be added on 1 if line ?!

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
eduardDudu: Can be added on 1 if line ?!

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
eduardDudu: Yes, you are right. Thank you

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
alexandru-calinoiu: Not really, one idea will be to extract this into specialized objects that will hold the validation logic.

coffeeshop/src/main/java/ro/eduarddudu/coffeeshop/CSVReader.kt
alexandru-calinoiu: One possible way is to use a data object that will have named properties.

alexandru-calinoiu:
Thank you, will followup via email.

PR #8, opened 2019-02-22
Solution to oop test

wokePanda:
# test_oop

This is a solution to the oop test. 

The program is supposed to take the user's location and a CSV file with all the coffee shops and provide a list of the 3 closest ones.

## Installation

Use [Homebrew](http://brew.sh/) to install Kotlin.

```bash
$ brew update
$ brew install kotlin
```

## Usage

The project can be run from the command line.

### Command line

In order to compile the project navigate to the project folder and run the following command:

```bash
$ kotlinc src -include-runtime -d test_oop.jar
```
This will create a file named: ``` test_oop.jar ```

In order to run the application type in the following command:

```bash
$ java -jar test_oop.jar <user x coordinate> <user y coordinate> <shop data filename>
```

Make sure that the ```.csv``` file is located in the project folder.


tests/Model/CoffeeShopTest.kt
```diff
@@ -0,0 +1,32 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+internal class CoffeeShopTest {
+
+    @Test
+    fun createCoffeeShop() {
+        val coffeeShop = CoffeeShop(Coordinates(42.4,143.234), "Starbucks")
+        assertEquals("Starbucks", coffeeShop.name)
```
alexandru-calinoiu: This assertion adds no value, you are testing that property setting working in kotlin, this is a language feature that the kotlin developers should have tested.

tests/Model/CoffeeShopTest.kt
```diff
@@ -0,0 +1,32 @@
...
+        assertEquals(42.4, coffeeShop.coordinates.latitude)
+        assertEquals(143.234, coffeeShop.coordinates.longitude)
```
alexandru-calinoiu: The above comments applies to this assertion also but there are a couple of other issues with it:
- you are testing the implementation of another class, it should not be the concern of this test
- this assertion can be done like `assertEquals(Coordinates(42.4, 146.234), coffeeShop.coordinates)` can you tell me why?

tests/Helpers/ConverterTest.kt
```diff
@@ -0,0 +1,50 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+internal class ConverterTest {
+
+    @Test
+    fun convertLineToCoffeeShop() {
```
alexandru-calinoiu: The naming of the tests is not that great, this guy has a nice [blog post](http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html) about it.

tests/Helpers/ConverterTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val starbucksNY = CoffeeShop(Coordinates(2.923489,-44.234128), "Starbucks New York")
+        val computedCoffeeShopsList = Converter().linesToCoffeeShops(lines)
+
+        assertEquals(starbucksRio.name, computedCoffeeShopsList[0].name)
+        assertEquals(starbucksSF.coordinates.latitude, computedCoffeeShopsList[1].coordinates.latitude)
+        assertEquals(starbucksNY.coordinates.longitude, computedCoffeeShopsList[2].coordinates.longitude)
+    }
+
+    @Test
+    fun convertListToString() {
```
alexandru-calinoiu: This is a very confusing test, not sure what is used for.

src/Model/CoffeeShop.kt
```diff
@@ -0,0 +1,14 @@
+
+internal data class CoffeeShop(val coordinates: Coordinates, val name: String) {
+    var distanceFromUser: Double? = null
```
alexandru-calinoiu: Mutability is considered by many not a good idea, can you tell me a couple of reasons why is that?

src/Helpers/DistanceUtility.kt
```diff
@@ -0,0 +1,15 @@
+import kotlin.math.pow
+import kotlin.math.sqrt
+
+internal class DistanceUtility {
+
+    fun calculateDistance(from: Coordinates, to: Coordinates): Double {
```
alexandru-calinoiu, line 6: This method should be made private, why do you think is a good idea to limit the public api of a class?

alexandru-calinoiu (commented):
This is a good start,

- oop solution is very basic, would have loved to see some stronger separation of concerns, layers like data (cvs reading), business (closes shop calculations) should have been separated by interfaces so they can easily be tested in isolation
- test are a mix of unit and integration test and have some issues

tests/Model/CoffeeShopTest.kt
wokePanda: Because the ```Coordintes``` class is a data class, meaning it has an autogenerated implementation of ```equals```.

tests/Helpers/ConverterTest.kt
wokePanda: Here I wanted to test the ```listToString``` function of the ```Converter``` class.

Basically I wanted to make sure that if I pass a list of ```CoffeeShop``` objects to this function, along with a number of top results (```closestCount```) I get a string with the format requested for the output of the program.

src/Model/CoffeeShop.kt
wokePanda: - It can lead to unintended side-effects;
- It's not thread safe. As the object can be changed, multiple threads might try to update it at the same time, or they might see separate versions of it;
- It is easier to think about immutable objects since you don't have to think about the ways they might change over time (this can help in debugging for example).

src/Helpers/DistanceUtility.kt
wokePanda: In the context of OOP it's a good idea since you can break up tasks into smaller parts. This way if you have a function that is used internally by the class (to serve another function) then you don't want other classes to have access to it (or to use it). An example of this could be a function that changes the value of an object inside the class, this function can be called inside the class by other functions (as an effect of some other event), but you don't want other classes to access it directly.

levelboy:
Already talked this thoruhg. Closing. Thank you

PR #9, opened 2019-02-24
My swift solution for the test

c-Viorel:


FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+            CoffeeShop.init(name: "Starbucks4", location: Coordinates.init(latitude: 4, longitude: 4)),
+        ]
+        
+        coffeeShopFinder   =  CoffeeShopFinder.init(coffeeShops, Coordinates.init(latitude: 0, longitude: 0))
+        closestCoffeeShops =  coffeeShopFinder.closestCoffeeShops
+    }
+    
+    // TODO: implement teardown
+    
+    public func test() {
```
alexandru-calinoiu, line 24: Test naming can be improved, I find this [blog post](http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html) helpfull.

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+        XCTAssertNotNil(coffeeShopFinder)
```
alexandru-calinoiu, line 25: This assertion is not very useful, it looks like it tests your test setup.

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+        XCTAssertNotNil(closestCoffeeShops)
+        XCTAssertEqual(closestCoffeeShops.count, 3)
```
alexandru-calinoiu, line 27: Usually multiple asserts are considered a bad practice can you, can you think why?

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+            CoffeeShop.init(name: "Starbucks2", location: Coordinates.init(latitude: 2, longitude: 2)),
+            CoffeeShop.init(name: "Starbucks3", location: Coordinates.init(latitude: 3, longitude: 3)),
+            CoffeeShop.init(name: "Starbucks4", location: Coordinates.init(latitude: 4, longitude: 4)),
+        ]
+        
+        coffeeShopFinder   =  CoffeeShopFinder.init(coffeeShops, Coordinates.init(latitude: 0, longitude: 0))
+        closestCoffeeShops =  coffeeShopFinder.closestCoffeeShops
+    }
+    
+    // TODO: implement teardown
```
alexandru-calinoiu, line 22: TODO?

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopLocationsParserTests.swift
```diff
@@ -0,0 +1,21 @@
...
+    
+    var coffeeShops:[CoffeeShop]?
+    
+    public override func setUp() {
+        super.setUp()
+        let input = "Starbucks Seattle,47.5809,-122.3160"
+        coffeeShops = try? CoffeeShopLocationsParser.parse(inputDataCSVString: input)
+    }
+    
+    public func test() {
```
alexandru-calinoiu, line 14: Data can be sparse, what about locations that don't have a name, or one of the coordinates, or have invalid coordinates. Can you think of ways the user still get's some meaningful output? If yes, have tests to prove it.

FindMyCoffee.playground/Sources/FindMyCoffeApp.swift
```diff
@@ -0,0 +1,54 @@
...
+    // MARK:- Constructor
+    
+    public init(_ userLocation:Coordinates, coffeeShopsDatabase:String) {
+        self.userLocation       = userLocation
+        self.coffeeShopsDatabase = coffeeShopsDatabase
+    }
+    
+    // MARK:- Public methods
+    
+    public func findSolution() -> [CoffeeShop] {
```
alexandru-calinoiu, line 21: I wonder if this method needs to be public? `showSolution` seems to be enough from the problem requirements. Can you please explain why is this public?

FindMyCoffee.playground/Sources/FindMyCoffeApp.swift
```diff
@@ -0,0 +1,54 @@
...
+            print(coffeeShop.formattedDescription)
+        }
+    }
+    
+    // MARK:- Private methods
+
+    private func parseCoffeeShopDatabase() -> [CoffeeShop]? {
+        do {
+            
+            return try CoffeeShopLocationsParser.parse(inputDataCSVString: coffeeShopsDatabase)
```
alexandru-calinoiu, line 43: The direction of this layers is not something that I expected, as I expected the list of shop's to be input so flow will be something along the lines:

input[coffeshoplist, userlocation] (presentation) -> validate (business) -> find closes (business) -> show (presentation)

Not sure why will you need to reach into a data layer for this particular app.

alexandru-calinoiu (commented):
This is a good start, I like the fact that classes are separated into files this shows sights of a good separation of concerns.
Testing is a little spotty, please review my comments.

FindMyCoffee.playground/Sources/FindMyCoffeApp.swift
c-Viorel: Yes, you are right. I had in mind the possibility of getting  the solution without calling the showSolution() method, in case I needed to process them in a different way (like parse in another format)

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
c-Viorel: More work for me to debug the problem. I am the one who needs to find which assert has failed. So in the end, yes it's highly recommended having as few as possible asserts. I should have one more separate test for the constructor.

FindMyCoffee.playground/Sources/FindMyCoffeApp.swift
c-Viorel: The name of the function can be too general. In my mind, any kind of persistent information is a database, regardless of it's format. So in this method, I've just called the parser and treat the exceptions.

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopLocationsParserTests.swift
c-Viorel: For the locations without a name, I can provide a default generic name, but if one of coordinates are wrong I can't think to a valid solution without affecting the precision of location. Also, if the input data is inconsistent I have that exception that is triggered.

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
c-Viorel: I've forgotten to remove it. Initially, I wanted to use tearDown() to set to nill closestCoffeeShops & coffeeShopFinder.

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopFinderTests.swift
c-Viorel: Yes, I'm gonna change the names

FindMyCoffee.playground/Sources/FindMyCoffeTests/CoffeeShopLocationsParserTests.swift
alexandru-calinoiu: You could also skip over bad records, and still over the user a solution from the valid ones, it won't be that nice but at least they can still grab some coffee.

c-Viorel: Instead of throwing an error I can use continue to skip over invalid data, and a solution is still provided.  
I'm not very pleased about doing that, i'd prefer a full error instead a half of the solution, but in some cases, it may be useful.

alexandru-calinoiu:
Closing this and will followup via email.

PR #10, opened 2019-03-27
Coffee Addict 101

alinschueller:


INeedCoffee/coffeeShop.rb
```diff
@@ -0,0 +1,25 @@
+require './calculate'
+require './import'
+require './display'
+require './sort'
+
+class CoffeeShop
+  def self.INeedCoffee()
+    userLatitude = Float(ARGV[0])
```
alexandru-calinoiu, line 8: If one does not pass the right params the app will crash with a very strange stacktrace:

```
	2: from coffeeShop.rb:25:in `<main>'
	1: from coffeeShop.rb:8:in `INeedCoffee'
coffeeShop.rb:8:in `Float': can't convert nil into Float (TypeError)
```

It will be great to display minimal usage information, like you have in other command line tools like `git`

INeedCoffee/coffeeShop.rb
```diff
@@ -0,0 +1,25 @@
...
+    userLongitude = Float(ARGV[1])
+    coffeeShopFileName = ARGV[2]
+
+    coffeeShopName = Array.new
+    coffeeShopLatitude = Array.new
+    coffeeShopLongitude = Array.new
+    distanceFromUserToCoffeeShop = Array.new
+    sortedDistanceFromUserToCoffeeShop = Array.new
+
+    Import.CSVDataIntoCoffeeShop(coffeeShopFileName, coffeeShopName, coffeeShopLatitude, coffeeShopLongitude)
```
alexandru-calinoiu, line 18: All this functions have side effects, mutating inputs will likely result in a very hard to maintain software.

Can you tell me why function with side effects are considered a bad practice, and what is the opposite of such a function?

alexandru-calinoiu (commented):
Pros:
- the variable names are very good 
- the solution seems to separated into logical steps, hints to [railway programming](https://fsharpforfunandprofit.com/rop/) but the implementation ... (see cons)

Cons:
- ruby is practically butchered, no coding conventions no language sugar ...
- [primitive obsesion](https://refactoring.guru/smells/primitive-obsession) no usage of higher level objects
- unit testing is mixed with integration testing and it's poorly done

Overall it feels like the solution was rushed. 
My vote will be to redo the test in a language that is more comfortable to the you or spend a couple of days in our office working on a more realistic problem with a more comfortable language.

INeedCoffee/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
drqCode, line 1: .idea files are for RubyMine or related IDEs. These are not needed in the repo. Add them to .gitignore.

INeedCoffee/calculate.rb
```diff
@@ -0,0 +1,25 @@
+require 'haversine'
+
+class Calculate
+  def self.DistanceBetweenCoffeeShopAndUser(coffeeShopName, coffeeShopLatitude,coffeeShopLongitude,userLatitude,userLongitude, distanceFromUserToCoffeeShop)
```
drqCode, line 4: Use [naming conventions for ruby](http://rubylearning.com/satishtalim/ruby_names.html)

INeedCoffee/display.rb
```diff
@@ -0,0 +1,14 @@
+class Display
```
drqCode, line 1: 'Display' isn't an usual name for what the class does.

INeedCoffee/sort.rb
```diff
@@ -0,0 +1,14 @@
+class Sort
+  def self.DistanceBetweenCoffeeShopAndUser(coffeeShopName, distanceFromUserToCoffeeShop, sortedDistanceFromUserToCoffeeShop)
```
drqCode, line 2: Pluralize the names accordingly. Search more on how to implement such logic in Ruby.

levelboy:
This is no longer needed. Will be redone in a different language.

PR #11, opened 2019-04-03
Coffee App

bogdanterzea:
Read location and file from console and show first 3 nearest locations and distance.

### DIFF
--- a/CoffeeApp.js
+++ b/CoffeeApp.js
@@ -1,82 +1,88 @@
-﻿const fs = require('fs'); 
-const csv = require('fast-csv');
+﻿"use strict";
+var LintStream = require('jslint').LintStream;
+var fs = require('fs');
+var csv = require('fast-csv');
 var obj = csv();
 
-const readline = require('readline');
-const rdl= readline.createInterface(process.stdin, process.stdout);
+var readline = require('readline');
+var rdl = readline.createInterface(process.stdin, process.stdout);
 
 var longi;
 var lati;
 var file;
 var latitude = [];
-let longitude = [];
-let location = [];
+var longitude = [];
+var location = [];
 var ct = 0;
 
-rdl.question("Datele: ", function(data){
-    let elementss = data.split(',');
+rdl.question("Datele: ", function (data) {
+    var elementss = data.split(','),
+        value,
+        lat1,
+        lon1,
+        lat2,
+        lon2;
     longi = elementss[0];
     lati = elementss[1];
     file = elementss[2];
-    let value;
 
     function interpretation(value) {
-        location=[...location, `${value[0]}`];
-        latitude=[...latitude, `${value[1]}`];
-        longitude=[...longitude, `${value[2]}`];
+        location = [...location, `${value[0]}`];
+        latitude = [...latitude, `${value[1]}`];
+        longitude = [...longitude, `${value[2]}`];
     }
 
-    let lat1, lon1, lat2, lon2;
-
-    function getDistance(lat1,lon1,lat2,lon2) {
-        var R = 6371; // Radius of the earth in km
-        var dLat = deg2rad(lat2-lat1);  // deg2rad below
-        var dLon = deg2rad(lon2-lon1); 
-        var a = 
-          Math.sin(dLat/2) * Math.sin(dLat/2) +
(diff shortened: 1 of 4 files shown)


alexandru-calinoiu:
@bogdanterzea we are waiting on you to push the test before we review this, thank you.

alexandru-calinoiu (changes requested):
Before we can start review please act on the following:

- please review style and code conventions for javascript, the formating of the file look preaty bad, your can use https://www.npmjs.com/package/jslint to help with this
- please include in the pr description some help on how to run this? node version? example of test run.

CoffeeApp.js
```diff
@@ -0,0 +1,98 @@
...
+        value,
+        lat1,
+        lon1,
+        lat2,
+        lon2;
+    longi = elementss[0];
+    lati = elementss[1];
+    file = elementss[2];
+
+    function interpretation(value) {
```
alexandru-calinoiu: The name of the function is not very relevant, you are basically parsing a row.

alexandru-calinoiu: The name of the function is not very relevant, you are basically parsing a row.

CoffeeApp.js
```diff
@@ -0,0 +1,98 @@
...
+        latitude = [...latitude, `${value[1]}`];
+        longitude = [...longitude, `${value[2]}`];
+    }
+
+    function deg2rad(deg) {
+        return deg * (Math.PI / 180);
+    }
+
+    function getDistance(lat1, lon1, lat2, lon2) {
+        var R = 6371,// Radius of the earth in km
```
alexandru-calinoiu: This is most likely correct,

But the readme states:

```
Assume all coordinates lie on a plane.
```

So this makes for overengineering, can you tell me why this is considered usually a bad practice?

CoffeeApp.js
```diff
@@ -0,0 +1,98 @@
...
+   function showLocations() {
+       var result = [],
+           result1 = [],
+           distance = [],
+           i;
+
+       for (i = 0; i < location.length; i++) {
+           distance= [...distance, getDistance(longi,lati,latitude[i],longitude[i])];
+       }
+       for (i = 0; i < location.length; i++) {
```
alexandru-calinoiu: I think this can be merged with the above loop

CoffeeApp.js
```diff
@@ -0,0 +1,98 @@
...
+           console.log(`${resultSorted[i]['location']}, ${resultSorted[i]['distance']}`);
+       }
+   }
+
+    fs.createReadStream(`${file}`)  
+        .pipe(csv())
+        .on('data', (row) => {
+            interpretation(row);
+            ct++;
+             if ( ct === 6 ) {
```
alexandru-calinoiu, line 81: Mutating global variables are considered a bad practice, can you tell me a couple of reasons why people think this as what are some alternatives?

CoffeeApp.test.js
```diff
@@ -0,0 +1,9 @@
+var LintStream = require('jslint').LintStream;
+var { data } = require('./CoffeeApp');
+
+test('Verifica Starbuksurile apropiate', () => {
```
alexandru-calinoiu, line 4: Can you tell me what is considered to be the optimal rate of test code vs production code?

alexandru-calinoiu (commented):
I have no doubt this works, 

Please take the time to reply to the comments I've left.

CoffeeApp.js
bogdanterzea: Overengineering is considered a bad practice because you make a solution more complicated than it needs to be to fit your needs

CoffeeApp.js
bogdanterzea: Mutating global variables are considered a bad practice because you can overwrite your variables if you have more functions that depend on them.

CoffeeApp.test.js
bogdanterzea: I don't quite understand the question.

alexandru-calinoiu: There a couple of [writings](http://wiki.c2.com/?ProductionCodeVsUnitTestsRatio) out there that suggest that the production code to test code ratio should be aroud 1:2, in other words for every line of production code you should have roughly 2 lines of test code.

alexandru-calinoiu:
Closing this, will followup via email.

PR #12, opened 2019-04-10
AgileFreaks OOP Test for Madalin Balint

MadalinBalint:


src/com/madalinb/coffeeshop/csv/CSVParser.kt
```diff
@@ -0,0 +1,31 @@
...
+                        params[0].trim(),
+                        params[1].toDouble(),
+                        params[2].toDouble()
+                    )
+                } else
+                    println("Error: not a valid X and/or Y coordinate in line '$line'")
+            } else
+                println("Error: not a valid name in line '$line'")
+        } else
+            println("Error: wrong number of parameters in line '$line'")
```
pirvudoru: What would it take to use this class in a web-based application instead of a console one?

src/com/madalinb/coffeeshop/abstractions/validations/ValidString.kt
```diff
@@ -0,0 +1,5 @@
+package com.madalinb.coffeeshop.abstractions.validations
+
+interface ValidString {
```
pirvudoru: By reading the name, I would think there will be a class that inherits from `String` and implements `ValidString`.

src/com/madalinb/coffeeshop/data/CSVFileOps.kt
```diff
@@ -0,0 +1,35 @@
+package com.madalinb.coffeeshop.data
+
+import com.madalinb.coffeeshop.abstractions.FileOperations
+import com.madalinb.coffeeshop.abstractions.Parser
+import java.io.File
+import java.io.FileNotFoundException
+import java.io.IOException
+
+class CSVFileOps(var parser: Parser<CoffeeShop>) : FileOperations<CoffeeShop>() {
```
pirvudoru: What does "CSVFileOps" stand for and the purpose of it?

.idea/libraries/junit_4_12.xml
```diff
@@ -0,0 +1,9 @@
+<component name="libraryTable">
```
pirvudoru, line 1: Why do we need `.idea` files added to VCS?

src/com/madalinb/coffeeshop/Main.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    if (coffeeShops.isEmpty()) {
+        println("Processing the file '$filename' has found no valid entries")
+        return
+    }
+
+    coffeeShops.forEach {
+        val distance = it.distanceTo(x, y)
+        map[distance] = it.name
+    }
```
pirvudoru: What can we do something about this logic if we want to re-use it in another context (web) ?

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
```diff
@@ -0,0 +1,7 @@
+package com.madalinb.coffeeshop.data
+
+data class CoffeeShop(var name: String, var x: Double, var y: Double) {
+    fun distanceTo(x2: Double, y2: Double): Double {
```
pirvudoru: Why does the method accept x and y? What alternatives do we have and how does it compare to this one?

src/com/madalinb/coffeeshop/tests/CoffeeShopTest.kt
```diff
@@ -0,0 +1,34 @@
...
+        coffeeShop = CoffeeShop("Starbucks Seattle", 47.5809, -122.3160)
+    }
+
+    @Test
+    fun distanceTo() {
+        assert(coffeeShop?.distanceTo(47.6, -122.4) == 0.0861441234211632)
+    }
+
+    @Test
+    fun getName() {
```
pirvudoru: By looking at the name of the test, i don't have much context of what it actually verifies. 

https://stackoverflow.com/questions/155436/unit-test-naming-best-practices

pirvudoru (commented):
Looks good! Have a few questions / suggestions that i'd like to discuss.

src/com/madalinb/coffeeshop/tests/CoffeeShopTest.kt
MadalinBalint: Changed the name to a more appropriate one.

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
MadalinBalint: In this case, because the documentation says that we have to assume that all coordinates lie on a plane, we are talking about Cartesian coordinates. In a real world where the Earth is not flat (as some people really belive), but round, the coordinates should be latitude and longitude, and the formula for calculating the distance would be different.

.idea/libraries/junit_4_12.xml
MadalinBalint: They were added by mistake. Any IDE related configuration files shouldn't be commited to the repository.

src/com/madalinb/coffeeshop/data/CSVFileOps.kt
MadalinBalint: File operations for the CSV file format. I've changed the name to a more appropriate one.

src/com/madalinb/coffeeshop/abstractions/validations/ValidString.kt
MadalinBalint: I've changed the name to a more appropriate one.

src/com/madalinb/coffeeshop/csv/CSVParser.kt
MadalinBalint: Use a different method to show the error messages (Console.log) or better throw them as an Exception.

src/com/madalinb/coffeeshop/Main.kt
MadalinBalint: Move the variable declaration outside of the loop.

src/com/madalinb/coffeeshop/Main.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    val map = mutableMapOf<Double, String>()
+    val df = DecimalFormat("#.####")
+    df.roundingMode = RoundingMode.CEILING
+
+    val x: Double? = args[0].toDoubleOrNull()
+    val y: Double? = args[1].toDoubleOrNull()
+    val filename = args[2]
+
+    if (args.size != 3) {
```
alexandru-calinoiu, line 13: Isn't this check to late, as you already have assigned the params in the above lines.

src/com/madalinb/coffeeshop/csv/CSVParser.kt
```diff
@@ -0,0 +1,31 @@
...
+    Parser<CoffeeShop> {
+    override fun parseLine(line: String): CoffeeShop? {
+        val params = line.split(",")
+
+        if (validator.hasEnoughArguments(params)) {
+            if (validator.isValidName(params[0])) {
+                if (validator.isValidDouble(params[1]) &&
+                    validator.isValidDouble(params[2])
+                ) {
+                    return CoffeeShop(
```
alexandru-calinoiu: This reminds me of https://www.audero.it/blog/wp-content/uploads/2016/08/php-street-fighter.jpg

I think this validator has a very large api surface, it should simply state if a line is valid or not.

Can you think of a better way to do this?

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
alexandru-calinoiu: I think @pirvudoru is refering here to a code smell called [primitive obsesion](https://refactoring.guru/smells/primitive-obsession) you are inferring that 2 doubles are coordinates, but we are asking if there is a better way to do this.

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
```diff
@@ -0,0 +1,7 @@
+package com.madalinb.coffeeshop.data
+
+data class CoffeeShop(var name: String, var x: Double, var y: Double) {
```
alexandru-calinoiu, line 3: Usually data classes are meant to store only data and no functions, can you think of why is the case?

src/com/madalinb/coffeeshop/Main.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    if (coffeeShops.isEmpty()) {
+        println("Processing the file '$filename' has found no valid entries")
+        return
+    }
+
+    var distance:Double
+    coffeeShops.forEach {
+        distance = it.distanceTo(x, y)
+        map[distance] = it.name
```
alexandru-calinoiu: This might not be correct, if you use a map and you have 2 shops at the same distance you will only show one to the user.

src/com/madalinb/coffeeshop/Main.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    val sortedMap = map.toSortedMap()
+    var counter = 0
+
+    for ((dist, name) in sortedMap) {
```
alexandru-calinoiu: Preaty sure there is a easier way to get the first 3 elements in kotlin

alexandru-calinoiu (commented):
Good job,

I like the fact that there where no dependencies on concrete classes, this makes the implementation easy to test.

Please take a few minutes to go over my comments and answer them.

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
MadalinBalint: Moved the function into a more appropriate class.

src/com/madalinb/coffeeshop/Main.kt
MadalinBalint: Moved the declarations after the check for the 3 arguments.

src/com/madalinb/coffeeshop/csv/CSVParser.kt
MadalinBalint: Refactored the class, and introduced Exceptions.

src/com/madalinb/coffeeshop/data/CoffeeShop.kt
MadalinBalint: Refactored the class.

src/com/madalinb/coffeeshop/Main.kt
MadalinBalint: Changed to List.

src/com/madalinb/coffeeshop/Main.kt
MadalinBalint: Refactored in class CoffeeShopFinder as function findClosest3Shops.

alexandru-calinoiu:
Thank you, will close this and followup via email.

PR #13, opened 2019-04-10
Added Solution

alinschuller:


alexandru-calinoiu (changes requested):
Please clean up the repo and only commit meaningful files, use a .gitignore generator, for example we don't need a `bin/` folder 

Also it will be helpfull to update the readme with minimal instructions on how to run this.

INeedCoffee.Tests/UserTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+            string[] coordinates = { "45.7838552", "24.1515281" };
+
+            UserCoordinatesCapture.Capture(userCoordinates, coordinates);
+
+            string actual = userCoordinates.Latitude.ToString() + " " + userCoordinates.Longitude.ToString();
+
+            string expected = "45.7838552 24.1515281";
+
+            Assert.Equal(expected, actual);
```
pirvudoru, line 22: Why did you choose to concatenate values and assert on that vs checking individiual properties?

INeedCoffee/CoffeeShop/CoffeeShop.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace INeedCoffee
+{
+    public class CoffeeShop
+    {
+        public string Name { get; set; }
+        public double Latitude { get; set; }
+        public double Longitude { get; set; }
```
pirvudoru, line 7: Seems that Latitude and Longitude always go together (ex: CoffeeShop and UserLocation). What can we do about them taking OOP into consideration?

INeedCoffee/CoffeeShop/CoffeeShop.cs
```diff
@@ -0,0 +1,10 @@
...
+        public double DistanceToUser { get; set; }
```
pirvudoru, line 8: Can a CoffeeShop exist without a distance? Be careful when mutating objects because it's going to become a mess if we loose track of where the mutations are made, which in turn will create unwanted, hard to debug behavior.

INeedCoffee/Utilities/HaversineDistanceCalculator.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace INeedCoffee
+{
+    public static class HaversineDistanceCalculator
+    {
+        public static double Calculate(double firstLatitude, double firstLongitude, double secondLatitude, double secondLongitude)
+        {
+            var earthRadiusInKm = 6372.8;
```
pirvudoru, line 9: Problem statement says:

```
Assume all coordinates lie on a plane.
```

INeedCoffee/Validation/CSVFilePathValidator.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.IO;
+
+namespace INeedCoffee
+{
+    public static class CSVFilePathValidator
+    {
+        public static bool Validate(string[] csvFileName)
```
pirvudoru, line 8: Can you name Pros / Cons of using bool as a return type? What about passing a bool as a parameter to a method?

INeedCoffee/Validation/TerminalArgumentsValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            bool coordinatesAreValid = UserCoordinatesValidator.Validate(terminalArguments);
+
+            if (coordinatesAreValid == false)
+            {
+                return false;
+            }
+
+            return true;
+            }
```
pirvudoru, line 31: File does not look formatted correctly. 

https://coryrylan.com/blog/why-enforcing-code-style-is-important

INeedCoffee/Validation/CSVFilePathValidator.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.IO;
+
+namespace INeedCoffee
+{
+    public static class CSVFilePathValidator
```
pirvudoru, line 6: What is the purpose of having multiple concrete Validators without the ability to interchange them? Would think this is not the pattern we want to use for our implementation. Probably you were looking for something like http://www.stavroskasidis.com/blog/2017/tips-and-tricks-1-guard-clauses/

INeedCoffee/CoffeeShop/ClosestCoffeeShopsFinder.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace INeedCoffee
+{
+    public static class ClosestCoffeeShopsFinder
+    {
+        public static void Find(List<CoffeeShop> coffeeShopsList, int numberOfClosestCoffeeShopsToShow, int numberOfDecimalsToRoundDistanceTo)
```
pirvudoru, line 9: Does not look too OOP. We are overusing static methods
https://stackoverflow.com/questions/2671496/java-when-to-use-static-methods

pirvudoru (changes requested):
Please answer the questions in my review :) Thanks!

INeedCoffee.Tests/UserTests.cs
alinschuller: When I built the test case, I based it on the from the UserCoordinatesCapture file, but definitely should’ve checked individual properties instead of being based on that method as I’ve done in UserCoordinatesValidator

INeedCoffee/CoffeeShop/CoffeeShop.cs
alinschuller: We can create an immutable struct for Coordinates

INeedCoffee/Utilities/HaversineDistanceCalculator.cs
alinschuller: In this case, we should return the following formula: 
- Math.Sqrt(Math.Pow((firstLatitude - secondLatitude), 2) + Math.Pow((firstLongitude - secondLongitude), 2);

INeedCoffee/CoffeeShop/CoffeeShop.cs
alinschuller: Yes, it definitely can exist without a distance and it makes sense it will cause issues if we want to scale up our application. I will make sure to document myself properly to avoid situations like this.

INeedCoffee/Validation/TerminalArgumentsValidator.cs
alinschuller: I formatted most of the files and most probably skipped this one.

INeedCoffee/Validation/CSVFilePathValidator.cs
alinschuller: Reading the tips and tricks from the link that you sent me and also reading the post from this link that I just found [(),](https://www.siepman.nl/blog/post/2014/01/05/SOLID-validation-class-with-easy-logical-implication.aspx) made me understand why my approach is not a good one and showed me a better one

alinschuller: Reading the tips and tricks from the link that you sent me and also reading the post from this link that I just found ([https://www.siepman.nl/blog/post/2014/01/05/SOLID-validation-class-with-easy-logical-implication.aspx](url)), made me understand why my approach is not a good one and showed me a better one

alinschuller: Reading the tips and tricks from the link that you sent me and also reading the post from this link that I just found ([https://www.siepman.nl/blog/post/2014/01/05/SOLID-validation-class-with-easy-logical-implication.aspx](https://www.siepman.nl/blog/post/2014/01/05/SOLID-validation-class-with-easy-logical-implication.aspx)), made me understand why my approach is not a good one and showed me a better one

INeedCoffee/CoffeeShop/ClosestCoffeeShopsFinder.cs
alinschuller: I had a moment in which I used the Visual Studio’s fix all utility without focusing on each method to define it properly

INeedCoffee/Validation/CSVFilePathValidator.cs
alinschuller: **Can you name Pros / Cons of using bool as a return type?**
PROS:
- Makes a function easier to read and evaluate
- Useful when a statement has only two possible outcomes 

CONS: 
- We cannot return any value other than True or False 
- Returning True or False does not give much information 

**What about passing a bool as a parameter to a method?**
Passing a bool as a parameter to a method can represent a code smell. One method can do only one thing and if it contains a bool parameter, most probably it means it does two things. 
Also, if we want to add in the future new conditions, this means that we will need to add new parameters which will make it very hard to maintain

alexandru-calinoiu:
Clone the project, run tests, and I get 3 failures

Screenshot from 2019-04-11 21-51-07

Most likely there is a setup step that I am missing.

INeedCoffee/INeedCoffee.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace INeedCoffee
+{
+    class INeedCoffee
```
alexandru-calinoiu, line 6: Is a funny class name, to bad that .net has this convention to prefix interfaces with `I`, so my instinct first time was what the ....

pirvudoru (approved):

INeedCoffee/Utilities/HaversineDistanceCalculator.cs
alexandru-calinoiu: You are right,

But what is the problem with your solution, given that you implemented haversine but problem stated something simpler? (it goes by a specific name and it's usually considered a bad thing)

alexandru-calinoiu (commented):
The solution get's the job done, is better that the initial solution, but it still suffers form tight coupling between classes, it has a bunch of other [code smells](https://blog.codinghorror.com/code-smells/) and the unit tests are rather sparse.

alinschuller:
Hi @alexandru-calinoiu , the failures occur because the file is not in that path that it specifies. I had the file initially in the Debug folder which I added in .gitignore and removed it through a couple of git commands. Therefore, the person who is missing a step is me, I should've referenced a path in which the file always existed and not only in the current folder. If you run the tests via terminal from the folder INeedCoffee_ReadyToRun, they should work because they take the file from the current folder.

INeedCoffee/Utilities/HaversineDistanceCalculator.cs
alinschuller: It's definitely the fact that I used static

INeedCoffee/INeedCoffee.cs
alinschuller: I realized that as well but somehow I liked the name so I just said "nah, I'll stick with it"

alinschuller: Yes, indeed, I just realized at a point that it looks like an Interface but I did not want to use MeWantCoffee, so I just said "nah, I'll stick with this even if it's weird"

alexandru-calinoiu:
Thank you, will close this and followup via email.

PR #14, opened 2019-07-26
CoffeeShop Solution

bogricc:
CoffeeShop o biblioteca PHP care incarca un fisier csv specificat si cauta pe baza coordonatelor primele 3 cele mai apropiate CoffeeShop-uri.


alexandru-calinoiu:
Before we can review this we need a couple of changes done:

- Please remove platform specific files from the project, choose one of the templates from https://github.com/github/gitignore
- Please update the pr description with how to run the project, imagine that the reviewer is not aware of any of the frameworks used it needs to know how to setup the project.

bogricc:
Updated readme, fixed commands
There are 2 variants to run and test the project.
I recommend Docker version.

alexandru-calinoiu:
@bogricc please do not include dependencies in your PR

See https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md

bogricc:
Ok. I am aware of this but the idea behind was to deliver a ready to run project.
But, I get the point. I will cleanup and adapt the readme file.

bogricc:
Fixed.
Things are clear now but, let me know if something is wrong.
Thank you!

Readme.md
```diff
@@ -39,3 +39,73 @@ Starbucks Seattle,0.0861
...
+
+To run all tests, please use this command:
+```php -f vendor/bin/phpunit -- --bootstrap ./vendor/autoload.php --testdox tests```
+Where `tests` is folder where tests are defined.
+If you need to test only one test then you need to append the desired class.
+Example:
+```php -f vendor/bin/phpunit -- --bootstrap ./vendor/autoload.php --testdox tests/CoffeeShopDistanceTest```
+
+
+### Within a Docker container
```
alexandru-calinoiu, line 67: Why not simply create the Dockerfile in the repo?

index.php
```diff
@@ -0,0 +1,24 @@
...
+php -f index.php 47.6 -122.4 coffee_shops.csv
+USG;
+
+if($argc !== 4){
+    echo $usage, " ($argc)", PHP_EOL;
+}
+else{
+    $coordX  = $argv[1];
+    $coordY  = $argv[2];
+    $csvFile = $argv[3];
```
alexandru-calinoiu, line 16: This variables have good names: coordX, coordY and csvFile clearly express their content.

index.php
```diff
@@ -0,0 +1,24 @@
...
+    try{
+        $v = new GetMyCoffee($coordX, $coordY, $csvFile);
+        $arr = $v->getNearShops();
```
alexandru-calinoiu: This variables don't have good names, can you improve them?

src/GetMyCoffee.php
```diff
@@ -0,0 +1,49 @@
...
+        //get the list of shops objs
+        $shopsList = $this->__reader->getList();
+
+        $this->__coffeeShops = new CoffeeShopsArray($shopsList);
+        foreach($this->__coffeeShops as $shop){
+            $shop->calcDistance($this->__userLocation);
+        }
+    }
+
+    public function getNearShops(){
```
alexandru-calinoiu, line 39: This is the only function that it is being used from the public api of the class, I think that the rest of them should no be public, can you explain why they where left public?

src/GetMyCoffee.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function setUserLocationByCoords($userXCoord, $userYCoord){
+        $userLocation = new Location($userXCoord, $userYCoord);
+        $this->__userLocation = $userLocation;
+        return $this;
+    }
+
+    public function process(){
+        //get the list of shops objs
```
alexandru-calinoiu: This comment is redundant, the naming is preatty good, so the comment does not bring any value.

src/GetMyCoffee.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+use CoffeeShop\ReadCoffeeShopList;
+use base\Location;
+use CoffeeShop\CoffeeShopsArray;
+
+final class GetMyCoffee{
```
alexandru-calinoiu, line 7: Naming classes with verbs make us think you wanted to use [command pattern](https://en.wikipedia.org/wiki/Command_pattern) but your implementation does not respect the pattern.

src/base/ReadCsvList.php
```diff
@@ -0,0 +1,52 @@
...
+    public function readCsv(){
+        $csv = array_map('str_getcsv', file($this->csvFile, FILE_SKIP_EMPTY_LINES|FILE_IGNORE_NEW_LINES));
+        return $csv;
+    }
+    public function setIndexes($csvHeader){
+        if(is_array($csvHeader)){
+            $headerLength = count($csvHeader);
+            $all = count($this->csvList, COUNT_RECURSIVE);
+            $lvl = count($this->csvList);
+            $csvColumnsSize = ($all-$lvl) / $lvl;
```
alexandru-calinoiu: This is clever, but does not really work for files like:

```
Starbucks Seattle,47.5809,-122.3160
Starbucks SF,37.5209
Sibiu, xxx,37.595242, 42
Starbucks Seattle2,47.5869,-122.3368
Starbucks Rio De Janeiro,-22.923489,-43.234418
Starbucks Sydney,-33.871843,151.206767
```

Is really hard to reason that you are actually trying to do a validation on format of the csv lines.

src/base/ReadCsvList.php
```diff
@@ -0,0 +1,52 @@
...
+                throw new \Exception("Header size does not match to csv column size.");
+            }
+            else{
+                $this->csvList = array_map(function($el) use ($csvHeader){
+                    return array_combine($csvHeader, $el);
+                }, $this->csvList);
+            }
+        }
+        else{
+            throw new \Exception("Csv Header is not an array.");
```
alexandru-calinoiu: Usually a good idea to [define your own exceptions](https://www.php.net/manual/en/language.exceptions.extending.php). Why do you think that is?

src/CoffeeShop/CoffeeShopDistance.php
```diff
@@ -0,0 +1,32 @@
+<?php
+namespace CoffeeShop;
+use base\ILocation;
+
+class CoffeeShopDistance extends CoffeeShop{
```
alexandru-calinoiu: Name for this was a little confusing, you probably meant `CoffeShopWithDistance`

src/CoffeeShop/CoffeeShopsArray.php
```diff
@@ -0,0 +1,24 @@
...
+        $list = $this->filterList($shopList);
+
+        parent::__construct($list);
+    }
+
+    public function filterList($list){
+        $newList = [];
+
+        $filterCB = function($item){
+            $v = ($item instanceof IShop);
```
alexandru-calinoiu, line 16: What else can it be in the $list param?

src/GetMyCoffee.php
```diff
@@ -0,0 +1,49 @@
...
+        return $this;
+    }
+
+    public function process(){
+        //get the list of shops objs
+        $shopsList = $this->__reader->getList();
+
+        $this->__coffeeShops = new CoffeeShopsArray($shopsList);
+        foreach($this->__coffeeShops as $shop){
+            $shop->calcDistance($this->__userLocation);
```
alexandru-calinoiu: The usage of `CoffeShopDistance` class is really strange, you have to initiate it and then call `calcDistance` for it be 'complete' and usable in the rest of the code.

If you don't know this implementation of this class is very tricky to use it.

src/CoffeeShop/CoffeeShop.php
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace CoffeeShop;
+use base\Location;
+
+class CoffeeShop extends Location implements IShop{
```
alexandru-calinoiu, line 5: When I think about a Coffee Shop I think `it is at a location` that `it is a location` the fact that it implements `ILocation` does not seem to help later on in the implementation.

src/CoffeeShop/CoffeeShop.php
```diff
@@ -0,0 +1,30 @@
...
+    private $__shopName;
+
+    public function __construct($xCoord, $yCoord, $shopName)
+    {
```
alexandru-calinoiu: The style is inconsistent, sometimes you use `{` on a new line sometime you put then on the same line.

tests/LocationTest.php
```diff
@@ -0,0 +1,33 @@
+<?php
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+use base\Location;
+
+final class LocationTest extends TestCase {
+
+    public function testCanBeCreatedWithZeroCoodinate(): void{
+        $this->assertInstanceOf(Location::class, new Location(0.0, 0.0) );
```
alexandru-calinoiu: One unplesant side effect of the constructor throwing errors is the fact that you now have to write this tests.

Reading this test makes me thing you are testing the fact that the constructor in php actually works.

tests/LocationTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->assertSame($y, $l->getYCoord());
+    }
+
+    public function testCannotBeCreatedWithAnInvalidCoordinateX(): void{
+        $this->expectExceptionMessage("Coordinate X is not valid.");
+        new Location("", 12.2);
+    }
+
+    public function testCannotBeCreatedWithAnInvalidCoordinateY(): void{
+        $this->expectExceptionMessage("Coordinate Y is not valid.");
```
alexandru-calinoiu: Assertions are based on text that makes the specs brittle, what if we decide to change this text or localize it?

See my comment about custom errors.

tests/ExpectedOutputTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+        $this->expectOutputString($expectedOutput);
+
+        $coordX = "47.6";
+        $coordY = "-122.4";
+        $csvFile = "coffee_shops.csv";
+
+        $v = new GetMyCoffee($coordX, $coordY, $csvFile);
+        $arr = $v->getNearShops();
+        CoffeeShop\PrintMyCoffee::console($arr);
```
alexandru-calinoiu: Is a good integretion test, not sure what is the benefit of asserting on the console output vs on asserting on the result array.

alexandru-calinoiu (commented):
Thank you for your submission:

- the setup with docker was smooth and nice, and the app seemed to return the right output on the first try

Please review our comments related to the implementation.

The tests don't seem to test a lot of the validation cases you have around the code.

Readme.md
bogricc: Yes, I can create it since we are seeing this repo like a testing project.

index.php
bogricc: I am always trying to improve myself.

index.php
bogricc: Development scars. Consider them fixed with proper names.

src/GetMyCoffee.php
bogricc: I left them public in the case if programmer would like to use them separately and do all the processing later, but I think this is a good idea, too.

src/GetMyCoffee.php
bogricc: Removed. Sometime intentionally I leave some comments which reminds me what should happen next.

src/GetMyCoffee.php
bogricc: Solved, except constructor :)

src/GetMyCoffee.php
bogricc: Actually not. Rather I wanted to write CoffeeNeeded, like a cliche.

src/base/ReadCsvList.php
bogricc: That is a statistical validation, and it's passing only if header length is equal with average row length. Indeed it's not effective in that case.
Maybe another math like harmonic mean would do the trick in some cases.
Here I wanted to do that validation before combining the header with each row from csv.
I am thinking to filter out the rows with wrong size and throw an Exception if filtering comes in effect.

src/base/ReadCsvList.php
bogricc: Exceptions work just fine which are better than nothing.
Always there is room for improvement.

src/CoffeeShop/CoffeeShopsArray.php
bogricc: In PHP each item from an array can be anything. So, here I'm trying to ensure that this array will contains only shops.

src/GetMyCoffee.php
bogricc: Yes, indeed this should be well documented.
As a resolution I would extend CoffeeShopsArray into a class that is aware of user location and get that calculation done at the creation time.

bogricc: I feel like are better ideas. Any suggestion?

src/CoffeeShop/CoffeeShop.php
bogricc: When I wrote this I was thinking that a Coffee Shop is a place/location which is obviously located somewhere which is actually a shop with a name.
Makes sense?

src/CoffeeShop/CoffeeShop.php
bogricc: I don't like to blame others for my mistakes.
My style is with ` { ` on the same line.
Sometime is happening that my editor wants to help me write faster. I should train it well.

tests/LocationTest.php
bogricc: In PHP empty string equals to zero if type is not taken into consideration.
Here I'm testing that locations with 0(zero) x y coordinates exists. In system with xOy coordinates this would be in the origin.

tests/LocationTest.php
bogricc: As you write above I should write my own Exceptions and assert well.
Another solution is to use code numbers for each error.

tests/ExpectedOutputTest.php
bogricc: Both unit test and integration tests should be done.
Asserting on the console ensures that the array is well interpreted and printed especially when encoding is taken into consideration.

src/GetMyCoffee.php
bogricc: Solved with integrating the user location into the object and calculating the distance into the distance getter.

bogricc:
Challenging review. I appreciate this.
Many open-points solved. So, please, take a look.

Thank you!

src/CoffeeShop/CoffeeShopsArray.php
alexandru-calinoiu: Ok, 

Ideally this check should not be required, can you think of a way to write the code so that this check is not required? (just describe it don't change your implementation)

Readme.md
alexandru-calinoiu: You won't usually create this Dockerfile in production project? Why is that?

src/CoffeeShop/CoffeeShop.php
alexandru-calinoiu: Not really,

Let's say later on you will extend this project with other functionality and will include online coffee shops, they surely aren't a phisical location.

What I am trying to ask is if you can think of alternatives to inheritance? What are some of the shortcoming of inheritance?

src/CoffeeShop/CoffeeShopException.php
```diff
@@ -0,0 +1,6 @@
+<?php
+namespace CoffeeShop;
+
+class CoffeeShopException extends base\BaseException {
```
alexandru-calinoiu, line 4: Adding a `CoffeeShopException` that it's used all over the place is no better then the general `Exception` that you had previously.

There are some benefits of having specific custom exceptions, for example if you have a `CoffeeShopCsvFileParseException` you can add a property to it name `lineNumber` that will contain the line number where the parse failed, allowing the enduser to easily figure out what the issue is.

Can you name some other benefits of using a custom exception?

src/base/ReadCsvList.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function readCsv(){
+        $csv = array_map('str_getcsv', file($this->csvFile, FILE_SKIP_EMPTY_LINES|FILE_IGNORE_NEW_LINES));
+        return $csv;
+    }
+    public function setIndexes($csvHeader){
+        if(is_array($csvHeader)){
+            $csvList = $this->csvList;
+            $csvHeaderSize = count($csvHeader);
+            $newCsvList = array_filter($csvList, function($el) use ($csvHeaderSize){
```
alexandru-calinoiu, line 36: Getting back the the point here.

If the file has a invalid row, something that we expect with user input, can you think of more user friendly ways to handle this?

src/CoffeeShop/CoffeeShopWithDistanceToUser.php
```diff
@@ -0,0 +1,48 @@
+<?php
+namespace CoffeeShop;
+use base\ILocation;
+
+class CoffeeShopWithDistanceToUser extends CoffeeShop{
```
alexandru-calinoiu, line 5: Why do you think this is an improvement?

src/CoffeeShop/CoffeeShopWithDistanceToUser.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private function setDistance($distance){
+        if(is_float($distance)){
+            $this->__distance = $distance;
+        }
+        return $this;
+    }
+
+    public function getDistance() : float{
```
alexandru-calinoiu, line 22: A getter that has side effects, usually not a good idea?

src/CoffeeShop/CoffeeShopWithDistanceToUser.php
```diff
@@ -0,0 +1,48 @@
...
+        $distance = 0.0;
+        if($this->__userLocation instanceof ILocation){
+            $distance = $this->calcDistance();
+        }
+        else{
+            throw new CoffeeShopException("User location not set. Cannot calculate and get distance.");
```
alexandru-calinoiu, line 28: Won't it be a better idea to throw this exception when the user location is set?

src/CoffeeShop/CoffeeShopWithDistanceToUser.php
```diff
@@ -0,0 +1,48 @@
...
+        return $distance;
+    }
+
+    public function setUserLocation(ILocation $userLocation){
+        if($userLocation instanceof ILocation){
+            $this->__userLocation = $userLocation;
+        }
+    }
+
+    public function calcDistance() : float{
```
alexandru-calinoiu, line 39: Why have this public?

alexandru-calinoiu (commented):
Added a couple of comments feel free to answer them.

I will followup in an email shortly.

Readme.md
bogricc: Not all projects are required to work in a container.

src/CoffeeShop/CoffeeShopsArray.php
bogricc: Want to comes into my mind right now is to instantiate each item of the array at the creation time.
The constructor get only the csv list as array.

src/CoffeeShop/CoffeeShop.php
bogricc: As an alternative the composition is used.
Into the CoffeeShop an attribute called location is placed.
One shortcoming is that it tight couples the implementation details between childs and parents.

src/CoffeeShop/CoffeeShopException.php
bogricc: What it comes to my mind is to catch different exceptions at different levels of the application.

src/CoffeeShop/CoffeeShopWithDistanceToUser.php
bogricc: This could be like a property from other languages?

PR #16, opened 2019-10-08
test_oop Ruby

Vali-Mandeal:


alexandru-calinoiu:
Please give a proper name to the pr and add a description to it with a minimal info on how to run the code, ruby version other deps ....

alexandru-calinoiu (changes requested):
Please give a proper name to the pr and add a description to it with a minimal info on how to run the code, ruby version other deps ....

Vali-Mandeal:
Added proper description

Main.rb
```diff
@@ -0,0 +1,19 @@
+dirPath = File.dirname(File.dirname(File.expand_path(__FILE__)))
+load dirPath + "/CoffeeShopApp/Objects/CoffeeShop.rb"
```
alexandru-calinoiu, line 2: When I try to run the code via

```
ruby Main.rb
```

I get the following error

        1: from Main.rb:2:in `<main>'
Main.rb:2:in `load': cannot load such file -- /<path>/CoffeeShopApp/Objects/CoffeeShop.rb (LoadError)
```

Please take a look at [require](https://ruby-doc.org/core-2.6.5/Kernel.html#method-i-require) and [require_relative](https://ruby-doc.org/core-2.6.5/Kernel.html#method-i-require_relative) kernel methods.

alexandru-calinoiu (changes requested):
Still very hard to review:

- Instruction on how to run should be in PR description not in a commit.
- See the comm that I left in the poorly named `Main.rb`

Every language has it's own idioms and coding styles, you have seemed to ignored them and this makes reviewing this impossible.

Please take a look at the https://www.rubocop.org/en/stable/ and focus on proper naming including filenames and variables.

alexandru-calinoiu:
Closing this for now, will followup via email.

PR #19, opened 2019-11-01
Flori bogoi test oop

FloriBogoi:


alexandru-calinoiu (changes requested):
Thank you for your pr,

There where 4 requirements for this:
- send the solution in a pr
- solve it in python
- write unit test to prove the functionality
- respect the SOLID principles

Out of this you only respected the first 2, the other 2 seems where totally ignored. Could you please explain the reasoning?

FloriBogoi:
For the Unit Tests, I misunderstood, I got only the part with testing and for that I did the duplicate check, the reason is that I never did before unit testing in python, my bad.
Regarding the SOLID principle, from what I got on the internet, one example was to use classes, which I did.

alexandru-calinoiu:
Thank you for your submission, will followup shortly via email.

PR #20, opened 2019-12-02
solution 1 and 2, see Solution_Description

icozma:


Solution_Description.md
```diff
@@ -0,0 +1,60 @@
...
+    read the shops, one by one 
+    calculate the distance to provided coordinates
+    insert the distance into a sorted structure, sol
+    remove last element from sol
+        if more than needed elements in solution
+    show/return sol
+
+config.py contains refs/links to:
+    a shop generator, in this instance a reader from csv
+    a distance calculator, this depends on coordinate system, here euclidian distance on a chartesian coordinate system;
```
alexandru-calinoiu, line 17: This might be useful, but the requirements stated that `Assume all coordinates lie on a plane.`

There is an entire debate if you should stick to the req or go above them and 'overdeliver', my personal opinion is that one should write minimal code that implements the req and it's easy to end extend.

In this case it should be obvious and very easy how to add a new distance calculation formula.

Solution_Description.md
```diff
@@ -0,0 +1,60 @@
...
+Implementation of shop generator and distance calculator can be found in the lib folder.
+
+The test folder contains the following cases: 
+    FindShops_test  it uses the provided csv sample and a case when there are two shops at the same distance
+    Location_test   checks if the euclidian distance is correctly calculated
+    ShopsCSV_test   checks if shops generator works on provided csv file 
+
+From console the program runs with the command:
+```
+python find_nearby_shops.py 47.6 -122.4  ./data/starbucks.csv
```
alexandru-calinoiu, line 32: Try to run this with python3 or 2 results in 

```
  File "find_nearby_shops.py", line 2, in <module>
    from sortedcontainers import SortedDict
ImportError: No module named sortedcontainers
```

Solution_Description.md
```diff
@@ -0,0 +1,60 @@
...
+```
+DELIMITER //
+CREATE DEFINER=`root`@`localhost` PROCEDURE `sp_nearest_shops`(
+	IN `ax` FLOAT,
+	IN `ay` FLOAT,
+	IN `n` INT  
+)
+    DETERMINISTIC
+BEGIN
+		SELECT shop_name, sqrt( POWER(ax-x, 2) + pow(ay-y, 2) ) as dist  FROM shops ORDER BY dist ASC LIMIT n;
```
alexandru-calinoiu, line 52: Cool, the good old sql :)

Solution_Description.md
alexandru-calinoiu: I had to first install the `sortedcontainers` module.

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+import config as config
+log = config.log
+gps = config.gps
+shops_generator = config.shops
+N = config.N
+
+''' check arguments 
+    iterates through shops and keep the closest config.N in sol
+'''
+def find_nearby_shops(args):
```
alexandru-calinoiu, line 13: Reading this function was straight forward and I could figure out what it does.

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+    try:
```
alexandru-calinoiu, line 14: I would think that parsing args should not be the responsibility of this function, 

If parsing and validation of input is done separately this function might not even get called.

Right now if I don't pass an argument

```
python3 find_nearby_shops.py 47.6  ./data/starbucks.csv 
````
I get a very strange error

```
ERROR:logger:Exception list index out of range 
<traceback object at 0x7facb35045f0>
ERROR:logger:Exception local variable 'file_name' referenced before assignment 
<traceback object at 0x7facb35045f0>
{}
```

Usually console apps will print out a usage of some sort if you don't pass the correct number of args.

File: find_nearby_shops.py [hunk 6]
```diff
@@ -0,0 +1,42 @@
...
+        a_coordinates = args[:-1]
```
alexandru-calinoiu, line 15: I think the name of this variable can be improved, if we look at the business requirements from the readme file can you come up with a name that makes it obvious what this variable stores?

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+    except Exception as e:
+        log.error("Exception {e} \n{trace}".format(e=e, trace=e.__traceback__))
+
+    ret = {}
+    try:
+        shops = shops_generator(file_name).get_shops()
+        sol = SortedDict()
+        
+        for shop in shops:
+            (shop_name, b_coordinates) = shop
```
alexandru-calinoiu, line 26: `b_coordinates` some comment as the other coordinates variable.

lib/Location.py
```diff
@@ -0,0 +1,26 @@
...
+    ''' From array of strings
+    '''
+    def __init__(self, coordinates):
+        try:
+            self.__coordinates = []
+            for c in coordinates:
+                self.__coordinates.append( float(c) )
+            self.__dim = len(self.__coordinates)
+        except Exception as e :
+            raise Exception("Coordinates format does not match!")
```
alexandru-calinoiu, line 17: I will think that exceptions can be part of your domain, can you think of any good reasons of using custom exceptions?

.vscode/settings.json
```diff
@@ -0,0 +1,13 @@
...
+        "-v",
+        "-s",
+        "./test",
+        "-p",
+        "*test.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.nosetestsEnabled": false,
+    "python.testing.unittestEnabled": true,
+    "python.pythonPath": "/Users/icozma/anaconda3/envs/cofeeshop/bin/python"
```
alexandru-calinoiu, line 12: This should not be in the repo

test/ShopsCSV_test.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+
+class TestShopsGeneratorMethods(unittest.TestCase):
+    def test_distance_metric(self):
```
alexandru-calinoiu, line 4: The test name looks misleading to me, this test looks like it's testing if parsing parses the shop names.

Solution_Description.md
pirvudoru, line 32: Executed this command and got an error

```
pirvudoru@work:~/Projects/AgileFreaks/test_oop$ python find_nearby_shops.py 47.6 -122.4  ./data/starbucks.csv
  File "find_nearby_shops.py", line 2, in <module>
    from sortedcontainers import SortedDict
ModuleNotFoundError: No module named 'sortedcontainers'
```

How can developers of a python app share the dependency list so that everyone can have the same version of packages installed?

.vscode/settings.json
pirvudoru, line 12: image

When trying to run from vscode. Are these settings supposed to be in the repository ?

lib/Location.py
```diff
@@ -0,0 +1,26 @@
...
+class Location:
+
+    ''' From array of strings
+    '''
+    def __init__(self, coordinates):
+        try:
+            self.__coordinates = []
+            for c in coordinates:
+                self.__coordinates.append( float(c) )
+            self.__dim = len(self.__coordinates)
```
pirvudoru, line 15: Why couple the constructor to an array of strings?
What is the benefit of using an array for (X,Y) coordinates ? (since problem states that the coordinates are on a plane)

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+def find_nearby_shops(args):
+    try:
+        a_coordinates = args[:-1]
+        file_name = args[2]
+    except Exception as e:
+        log.error("Exception {e} \n{trace}".format(e=e, trace=e.__traceback__))
+
+    ret = {}
+    try:
+        shops = shops_generator(file_name).get_shops()
```
pirvudoru, line 22: shops_generator -> shops
shops -> ShopsCSV

shops_generator#get_shops
shops#get_shops
ShopsCSV#get_shops

I don't know how to refer to the things above. Naming is inconsistent and I need to do aliases in my head to figure out what it is.

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+    except Exception as e:
+        log.error("Exception {e} \n{trace}".format(e=e, trace=e.__traceback__))
+
+    return ret
+        
+
+if __name__ == '__main__':
+    args = sys.argv[1:]
+    sol = find_nearby_shops(args) 
+    print(sol)
```
pirvudoru, line 42: From README:
```
Expected output

Starbucks Seattle2,0.0645
Starbucks Seattle,0.0861
Starbucks SF,10.0793
```

pirvudoru: Program return code is 0 on both ok / invalid input.

http://tldp.org/LDP/abs/html/exitcodes.html

find_nearby_shops.py
pirvudoru, line 26: Why not use a Shop as a class (following the pattern you have on "Location") ?

find_nearby_shops.py
```diff
@@ -0,0 +1,42 @@
...
+        log.error("Exception {e} \n{trace}".format(e=e, trace=e.__traceback__))
+
+    ret = {}
+    try:
+        shops = shops_generator(file_name).get_shops()
+        sol = SortedDict()
+        
+        for shop in shops:
+            (shop_name, b_coordinates) = shop 
+            d = gps( b_coordinates ).dist_to( gps(a_coordinates) )
```
pirvudoru, line 27: Have to do aliases so that i know that b_coordinates = shop_coordinates and a_coordinates = user_coordinates.

test/FindShops_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+class TestFindShopsMethods(unittest.TestCase):
+
+    '''  on provided data 
+    '''
+    def test_nearby_shops(self):
+        from find_nearby_shops import find_nearby_shops
+        sol = find_nearby_shops([47.6, -122.4, './data/starbucks.csv']) 
+        expected= {'Starbucks Seattle2': 0.0645, 'Starbucks Seattle': 0.0861, 'Starbucks SF': 10.0793}
+        self.assertDictEqual ( sol, expected , '...shall be EQ as per https://github.com/Agilefreaks/test_oop')
```
pirvudoru, line 11: :+1: I like that the tests test the actual behavior of the components.

Solution_Description.md
```diff
@@ -0,0 +1,60 @@
...
+
+### First solution
+
+One solution is pretty strighforward:
+    read the shops, one by one 
+    calculate the distance to provided coordinates
+    insert the distance into a sorted structure, sol
+    remove last element from sol
+        if more than needed elements in solution
+    show/return sol
```
pirvudoru, line 13: This is how the `find_nearby_shops` method should look like from the high level. There would have been no need for this documentation if it was merged with code.

lib/ShopsCSV.py
```diff
@@ -0,0 +1,47 @@
...
+        except Exception as e:
+            raise Exception("Format exception") 
+
+
+    ''' 
+        Iterates through the text file
+        Params: 
+            line_parser - interprets the line returning shop_name and an array of coordinates 
+    '''    
+    def get_shops(self, line_parser= parse_line):
```
pirvudoru, line 40: Why have `line_parser` as a method parameter and not used with other than the default value?

pirvudoru (commented):
Thanks!

I have some comments/questions and want to hear your thoughts.

alexandru-calinoiu (commented):
The solution does it job (both of them actually), I left some comments.

One general note that I have it would have been nice to have better commit messages to follow the thought process.

alexandru-calinoiu:
Closing this as there will unfortunately be no conversation.

PR #21, opened 2020-02-08
A .NET Core solution

andreidraghiciu:
### **The CoffeeNation app:**
I've added a .NET Core solution to the problem to be solved. I followed as much as I could the Clean Architecture and SOLID principles so that the app would be easily maintainable.

### **Prerequisites:**
In order to be able to build the solution, run tests and then run the app itself, you need first of all to make sure you have installed the .NET Core SDK, version  3.1.101 - this includes the runtime.
Please use this link [.NET Core SDK version 3.1.101](https://dotnet.microsoft.com/download/dotnet-core/thank-you/sdk-3.1.101-windows-x64-installer) in order to download the x64 version of the Windows installer, or you can use a [scripted installation](https://dotnet.microsoft.com/download/dotnet-core/scripts) if it suits better.

### **Building and running tests:**
In order to build the app and run the tests, I have included in the repository root folder a batch file, **run_build.bat** which will build the solution and output the result to the **_BuildOutput** folder. 

After the build is complete, the test run will start automatically and output test and code coverage results in the **_BuildReports** folder. Besides the info in the console, the tests run report folder and tests coverage report page should open automatically.

Also, an example run of the application will be triggered, using the example values as input.

### **Running the application:**
An example for the application run is included in **run_example.bat**, in the repository root, which will run using the example input parameters.

The application executable can be found in **_BuildOutput\CoffeeNation.ConsoleApp.exe**. Please note that the path to the .CSV file that is provided as input along with the user's coordinates, is relative to the current (working) directory.

Please let me know your thoughts. If you encounter any issues, please let me know. Many thanks!


src/CoffeeNation.ConsoleApp/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+using CoffeeNation.Dependency;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace CoffeeNation.ConsoleApp
+{
+    [ExcludeFromCodeCoverage]
+    class Program
+    {
+        [ExcludeFromCodeCoverage]
+        static async Task Main(string[] args)
```
pirvudoru, line 12: Haven't done C# in a while. It would help a lot to have a PR description or a README to instruct how to run the App / Tests.

src/CoffeeNation.Dependency/DependencyExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+            services.AddServiceRegistrations();
+            services.AddRepositoryRegistrations();
+            services.AddDataRegistrations(arguments);
+
+            return services;
+        }
+
+        private static void AddCoreRegistrations(this IServiceCollection services)
+        {
+            services.AddTransient<IDistanceCalculator, CartesianDistanceCalculator>();
```
pirvudoru: What is the benefit of registering the distance calculator in Transient scope over other scopes for the IDistanceCalculator ?

https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicelifetime?view=dotnet-plat-ext-3.1#Microsoft_Extensions_DependencyInjection_ServiceLifetime_Transient

tests/CoffeeNation.UnitTestsCommon/MockValues.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+
+namespace CoffeeNation.UnitTestsCommon
+{
+    public static class MockValues
```
pirvudoru, line 5: What is the difference between MockValues and MockData?

tests/CoffeeNation.UnitTestsCommon/MockData.cs
```diff
@@ -0,0 +1,197 @@
...
+        public static string ValidCsvLine3 => "Starbucks Moscow,55.752047,37.595242";
+
+        public static string ValidCsvLine99 => "Starbucks Seattle,47.5809,-122.3160";
+
+        public static IEnumerable<string> ValidCsvContent => new List<string>
+        {
+            ValidCsvLine1,
+            ValidCsvLine2,
+            ValidCsvLine3
+        };
```
pirvudoru: What if we need more than 3 lines and different values? Trying to understand why this file would be needed. It will only grow in size. It's like an `Utils` namespace.

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+    public class MessagingServiceTests
+    {
+        [Fact]
+        public async Task TestThat_SendErrorDetails_When_MessageIsNull_Throws_ArgumentOutOfRangeException()
+        {
+            // Arrange
+            var outputMessageRepositoryMock = new Mock<IOutputMessageRepository>();
+
+            var messagingService = new MessagingService(outputMessageRepositoryMock.Object);
```
pirvudoru, line 18: This setup is quite repetitive. Why is it preferred over a before each test setup?

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            // Act
+            async Task Act() => await messagingService.SendErrorDetails(MockData.NullMessage, MockData.GenericException);
+
+            // Assert
```
pirvudoru, line 23: Why do you prefer to add the Arrange / Act / Assert on each test?

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            await Assert.ThrowsAsync<ArgumentOutOfRangeException>(Act);
+        }
+
+        [Fact]
+        public async Task TestThat_SendErrorDetails_When_MessageIsEmpty_Throws_ArgumentOutOfRangeException()
```
pirvudoru, line 28: :+1: for test naming convention

src/CoffeeNation.Service/CoffeeShopsQueryService.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public CoffeeShopsQueryService(ICoffeeShopsQueryFacade queryFacade)
+        {
+            _userLocationRepository = queryFacade.UserLocationRepository;
+            _coffeeShopLocationRepository = queryFacade.CoffeeShopLocationRepository;
+            _distanceCalculator = queryFacade.DistanceCalculator;
+            _distanceSelector = queryFacade.DistanceSelector;
+        }
+
+        public async Task<IEnumerable<Distance>> GetClosestCoffeeShops()
```
pirvudoru: I would expect `GetClosesCoffeeShops` to return a list coffee shops (in terms of this App, instances of `Location`), not a distances.

src/CoffeeNation.Service/Facade/CoffeeShopsQueryFacade.cs
```diff
@@ -0,0 +1,23 @@
...
+using CoffeeNation.Repository.Interfaces;
+using CoffeeNation.Service.Interfaces.Facade;
+
+namespace CoffeeNation.Service.Facade
+{
+    public class CoffeeShopsQueryFacade : ICoffeeShopsQueryFacade
+    {
+        public IUserLocationRepository UserLocationRepository { get; }
+        public ICoffeeShopLocationRepository CoffeeShopLocationRepository { get; }
+        public IDistanceCalculator DistanceCalculator { get; }
```
pirvudoru, line 12: Why do we need to expose this on a Facade ? Afaik, the Facade should exists to expose the use cases, and not internal services used to accomplish the usecases.

src/CoffeeNation.Repository/UserLocationRepository.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        private readonly IUserLocationDataReader _dataReader;
+        public UserLocationRepository(IUserLocationDataReader dataReader)
+        {
+            _dataReader = dataReader;
+        }
+
+        public async Task<Location> GetUserLocation()
+        {
+            return await _dataReader.ReadUserLocation();
```
pirvudoru, line 18: Please help me understand the abstractions used and what is the responsability of each abstraction
- Repository
- Data Reader/Writer
- Input / Output Provider

From what i understand, the Repository delegates to the Data Reader/Writer
The data Reader/Writer does the serialization/deserialization of objects from/to a raw I/O (CsvContentProvider / UserLocationProvider). These are concrete per each repository.
The provider uses native apis and is the edge of the system.

src/CoffeeNation.Service/CoffeeShopsQueryService.cs
pirvudoru: Suppose we get another requirement to get the closest coffee shops to the AgileFreak's office. How would the system accomodate this requirement (what components would need to change?)

pirvudoru:
Everything looks thought out. Please take some time to answer the questions above.
Thank you!

src/CoffeeNation.ConsoleApp/Program.cs
andreidraghiciu: You are absolutely right, apologies for not being thorough with the PR description. I've updated the description and also have included a batch file that should help with the build and tests run and another one that runs the app as an example.

src/CoffeeNation.Dependency/DependencyExtensions.cs
andreidraghiciu: You are right, in this case there is no real benefit in using a transient scope. Due to the stateless behaviour of the DistanceCalculator class, it could have been as well singleton. Will update code shortly.

tests/CoffeeNation.UnitTestsCommon/MockValues.cs
andreidraghiciu: MockValues was intended to keep primitive types, values that can be reused to populate the fields of objects in MockData. MockData contains higher level / complex objects that are used through the app. A more suitable name for MockData would be MockObjects - the MockData name is a habit from modeling db objects I guess :) Will update to MockObjects.

tests/CoffeeNation.UnitTestsCommon/MockData.cs
andreidraghiciu: Agreed, the file will grow as the app grows and we can split it into partial classes when we need to, or even different classes if we feel the need to. Its purpose is to have all the mocked data into one place so that is easier to model the context for tests, in case objects that are somehow dependent need to be passed. It should help with giving meaning to the test cases not having magic values.

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
andreidraghiciu: In this and similar cases I didn't find it to be an overhead, I prefer to see all what the test does in one shot. In other instances, such as 'CoffeeShopsQueryServiceTests' the test methods would have been way to large, so I moved the default setup in the constructor.

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
andreidraghiciu: I believe I got used to this, in a test, it should be very clear where each of these stages take place and they should not be mixed. It also helps visually.

tests/CoffeeNation.Service.UnitTests/MessagingServiceTests.cs
andreidraghiciu: Tnanks! :) I tried different flavours until I got to this one.

src/CoffeeNation.Service/CoffeeShopsQueryService.cs
andreidraghiciu: I renamed the GetClosestCoffeeShops method to GetShortestCoffeeShopDistances as it is more relevant.

In order to be able to query distances for different user locations (such as AgileFreak's office) we need to change the UserLocationDataReader to allow it to receive multiple locations (different providers, multiple instances, can have different implementations) as long as we know the meaning of each location. We can then return an array of Location as user locations of different types. We can then change the repository method to allow filtering by Tag, or we can add a "type" property to the Location class to help out with filtering criteria. If we need to display distances from different starting points (user locations), it's best we add the filtering parameter to the service query method as well and pass it further to the repository.

src/CoffeeNation.Service/Facade/CoffeeShopsQueryFacade.cs
andreidraghiciu: Well, there were too many parameters sent to the constructor and the code became difficult to read and test. As a rule that I use, if parameters count passed via the constructor is greater than 3, then we need to group. A facade should be in order anytime we need to interact with multiple functionalities from different components.

src/CoffeeNation.Repository/UserLocationRepository.cs
andreidraghiciu: Everything you said above is exactly so. The providers are the edge of the system, they allow interaction with hard dependencies and also enable us to fake the interactions with the outside of the system. The readers/writers convert the input/output to concepts we can use to operate with, exactly like you said, serialization/deserialization and the repository allows us to define common selection criteria for the deserialized objects, so that the service layer can apply application/business logic (core), without the concern (and without the need to know how to) of querying an entire data set.

andreidraghiciu: I've attached a very basic application flow diagram
coffee_nation_diagram

andreidraghiciu:
I added the requested changes and tried to answer all questions. Please let me know if there are any other issues. 
Many thanks!

alexandru-calinoiu:
Thanks, will get back to you in an email soon.

PR #22, opened 2020-05-19
Ruby Solution

beniaminmuresan:
For solving this challenge I used `ruby 2.7.1` and `bundler 2.1.4`. See Gemfile. Run `bundle install` before running the main file.

How to run the program:
`ruby main.rb` - then give the input.
How to run all the tests:
`rspec spec/`
How to run individual test:
`rspec spec/distance_calculator_spec.rb`


user_coffee_shop_query.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class UserCoffeeShopQuery
+  attr_reader :distances
+
+  def initialize(coffee_shops, user)
+    @coffee_shops = coffee_shops
+    @user = user
+  end
+
+  def calculate_distances
```
drqCode: If this intended to be object internal method, it should be marked as `private`

spec/user_coffee_shop_query_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+RSpec.describe UserCoffeeShopQuery do
+  let (:filename) { 'coffee_shops.csv' }
+  let (:coffee_shops) { CoffeeShop.new(filename).all }
+  let (:user) { User.new(12, 43) }
+  let(:coffee_shops) { [{name: "Starbucks Seattle", x: "47.5809", y: "-122.3160"},
+                       {name: "Starbucks SF", x: "37.5209", y: "-122.3340"},
+                       {name: "Starbucks Moscow", x: "55.752047", y: "37.595242"},
+                       {name: "Starbucks Seattle2", x: "47.5869", y: "-122.3368"}] }
+
+  describe "#calculate_distances?" do
```
drqCode: This method is not used anywhere in the project other than in its internals. Shouldn't be better to test the public `#show_closest`?

drqCode (changes requested):
I don't agree with designing domain/problem object APIs for testing purposes. Maybe you can find it here to understand why I don't think is a good idea: https://semaphoreci.com/community/tutorials/getting-started-with-rspec

drqCode:
Everything else is :1st_place_medal:

user_coffee_shop_query.rb
beniaminmuresan: Absolutely right. My bad :+1:

spec/user_coffee_shop_query_spec.rb
beniaminmuresan: It makes much more sense this way. Fixed it. :+1:

spec/input_validator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:coffee_shops) { [["Starbucks Seattle", "47.5809", "-122.3160"],
+                       ["Starbucks SF", "37.5209", "-122.3340"],
+                       ["Starbucks Moscow", "55.752047", "37.595242"],
+                       ["Starbucks Seattle2", "47.5869", "-122.3368"]] }
+
+  describe '#validates' do
+    context 'when invalid input' do
+      it 'throws error and program exits' do
+        input_validator = InputValidator.new(invalid_input)
+        expect { input_validator.validate }.to raise_error
```
pirvudoru: Getting warnings about this.

spec/solution_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      expect(solution.length).to be(3)
+    end
+  end
+
+  describe '#validate_input' do
+    it 'forces the program to stop if the input is not valid' do
+      solution = Solution.new(input)
+      expect { solution.validate_input }.not_to raise_error
+      solution = Solution.new(wrong_input)
+      expect { solution.validate_input }.to raise_error(RuntimeError)
```
pirvudoru: Why did you choose to do asserts for different inputs in the same spec?

solution.rb
```diff
@@ -0,0 +1,36 @@
...
+    @user_y_coordinate = input[1].to_f
+    @csv_file_name = input[2]
+  end
+
+  def find
+    validate_input
+    user_coffee_shop_query.show_closest
+  end
+
+  def validate_input
```
pirvudoru: Seeing that `validate_input` is not part of the public API of the class. It's called only internally.

solution.rb
```diff
@@ -0,0 +1,36 @@
...
+  def user_coffee_shop_query
+    UserCoffeeShopQuery.new(coffee_shop.all, user)
+  end
+
+  def coffee_shop
+    CoffeeShop.new(csv_file_name)
+  end
+
+  def user
+    User.new(user_x_coordinate, user_y_coordinate)
```
pirvudoru, line 40: :+1:

coffee_shop.rb
```diff
@@ -0,0 +1,11 @@
+require 'csv'
+
+class CoffeeShop
+  def initialize(csv_file_name)
+    @csv_file_name = csv_file_name
+  end
+
+  def all
```
pirvudoru: I'm probably biased by ActiveRecord here, but was expecting this to be defined on the class.

pirvudoru: What is the reasoning for it returning an array of hashes?

coffee_shop.rb
```diff
@@ -0,0 +1,11 @@
+require 'csv'
+
+class CoffeeShop
+  def initialize(csv_file_name)
+    @csv_file_name = csv_file_name
```
pirvudoru: I think we are missing something here since one CoffeShop is backed by an entry in the CSV file, not by the whole file.

Understanding that the responsibility of this class is to parse the CSV and assuming the naming is not right.

user_coffee_shop_query.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class UserCoffeeShopQuery
+  attr_reader :distances
+
+  def initialize(coffee_shops, user)
+    @coffee_shops = coffee_shops
+    @user = user
+  end
+
+  def show_closest
```
pirvudoru: Why did you choose to add a public `show_closest` method on a query object ?

user_coffee_shop_query.rb
```diff
@@ -0,0 +1,26 @@
...
+    calculate_distances
+    @distances.sort_by { |coffee_shop| coffee_shop[:distance] }.first(3).each do |coffee_shop|
+      puts "#{coffee_shop[:name]},#{coffee_shop[:distance]}"
```
pirvudoru: The functionality of showing the first 3 items is highly dependent on the I/O method. How would the architecture change if we needed the same logic but export on an endpoint or write to a file?

input_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+  rescue
+    false
+  end
+
+  def valid_user_coordinates?
+    x_coordinate.numeric? && y_coordinate.numeric?
+  end
+
+  def valid_file_entries?
+    CSV.read(filename).each do |entry|
```
pirvudoru: There are 3 CSV.read in different places. The implementation is highly coupled and very hard to change / extend. Also reading a file 3 times is not so efficient.

pirvudoru (commented):
Hello. The solution works as expected. Would've been nice to have a PR description with instructions how to run app / tests.
Left some comments / questions. Please review.

user_coffee_shop_query.rb
beniaminmuresan: If we we're to send a string to an endpoint, I would first concatenate the whole answer and then send it to the endpoint/write to a file. Of course, a better option would be to send the instance variable distances which could be converted to json, being an array of hashes.

user_coffee_shop_query.rb
beniaminmuresan: You're right. It doesn't really make sense. Do you think maybe renaming to `result` is a better option?

coffee_shop.rb
beniaminmuresan: I think my mistake was that I was trying so hard to simulate a class that inherits from ActiveRecord, but here it wasn't the case. I will refactor this so it doesn't need a param `csv_file_name`.

coffee_shop.rb
beniaminmuresan: Firstly, CoffeeShop.all sounds a lot better(ActiveRecord like), thanks for the hint. Secondly, I was trying to make it more readable in other classes, that is the reason i converted to hash, but I'm not sure anymore how much does that help readability.

spec/input_validator_spec.rb
beniaminmuresan: Shoud've specificy the error class. Fixed it

spec/solution_spec.rb
beniaminmuresan: I should't have written this test at all. Just made `validate_input`a private method.

coffee_shop.rb
alex-damian-negru: Have you considered mapping to an entity/model/value instead?

input_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def valid_input?
+    valid_file? && valid_user_coordinates? && valid_file_entries?
+  end
+
+  def valid_file?
+    CSV.read(filename)
+  rescue
+    false
+  end
```
alex-damian-negru: This method seems to check if the CSV can be read, instead of checking if the file is valid (thinking `File.file?(filename)`). It's a bit more specific than the name suggests.

spec/coffee_shop_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  let(:filename) { 'coffee_shops.csv' }
+  let(:coffee_shops) { [["Starbucks Seattle", "47.5809", "-122.3160"],
+                       ["Starbucks SF", "37.5209", "-122.3340"],
+                       ["Starbucks Moscow", "55.752047", "37.595242"],
+                       ["Starbucks Seattle2", "47.5869", "-122.3368"]] }
+  describe "#all" do
+    before { allow(CSV).to receive(:read).with(filename).and_return(coffee_shops) }
+
+    it 'returns an array with all the coffee shops existing in the csv file' do
+      coffee_shop = CoffeeShop.new(filename)
```
alex-damian-negru: Entities instantiations like these can be extracted to a `let` instead of always defining them within the specs. It looks something like this:

```ruby
let(:coffee_shop) { CoffeeShop.new(filename) }
```

It will use the filename defined in another `let` so you can only leave the `expect`. Same in other specs.

spec/string_spec.rb
```diff
@@ -0,0 +1,14 @@
+require_relative '../string'
+
+RSpec.describe String do
+
+  describe "#numeric?" do
+    it 'determines if a certain string containes only numeric values' do
```
alex-damian-negru: A test should have one expectation. Consider using something like https://github.com/rspec/rspec-its for cases like this or extracting to separate `it` tests.

beniaminmuresan:
@drqCode, @pirvudoru, @alex-damian-negru Thanks for taking the time to review my solution. I tried to take into consideration every good advice you guys gave me. Of course, I did some stupid, rookie mistakes too, I tend to be nervous in this kind of situations. I think this version is way better.

user_coffee_shop_query.rb
drqCode: I knew it was just unintentional mistake:)

drqCode (approved):

user_coffee_shop_query.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class UserCoffeeShopQuery
+  def initialize(coffee_shops, user)
+    @coffee_shops = coffee_shops
+    @user = user
+  end
+
+  def result
+    distances.sort_by { |coffee_shop| coffee_shop[:distance] }.first(3)
+  end
```
drqCode, line 11: As a bonus challenge (only if you like challenges and you have time). Could you optimize this algorithm, IF it can be optimized better, for a huge list of coffee shops. At least more optimized, not really the best solution

user_coffee_shop_query.rb
```diff
@@ -0,0 +1,22 @@
...
+  private
+
+  def distances
+    @coffee_shops.map { |coffee_shop| { name: coffee_shop.name, distance: distance(coffee_shop) } }
+  end
+
+  def distance(coffee_shop)
+    DistanceCalculator.calculate(@user, coffee_shop)
+  end
+end
```
drqCode, line 24: As a bonus challenge (only if you like challenges and you have time). Could you optimize this algorithm, IF it can be optimized better, for a huge list of coffee shops. At least more optimized, not really the best solution

beniaminmuresan: I don't know if I know what you mean. Do you think that if the `calculate` method would receive all `coffee_shops` at once, this way we wouldn't instantiate a new object for each calculation. What do you think?

alex-damian-negru:
To upgrade your commit messages check out https://chris.beams.io/posts/git-commit/

spec/string_spec.rb
```diff
@@ -3,12 +3,30 @@
...
 
   describe "#numeric?" do
-    it 'determines if a certain string containes only numeric values' do
-      expect('123.123'.numeric?).to be true
-      expect('7654'.numeric?).to be true
-      expect('abc'.numeric?).to be false
-      expect('1a1.2'.numeric?).to be false
-      expect(''.numeric?).to be false
+    context 'when valid float number is given' do
+      it 'should return true' do
```
alex-damian-negru: You're almost there. :medal_sports: 

There's no need to wrap each spec in its own context. Using the context description as the spec description will bring it closer to what a good test should look like, provided best practices like http://www.betterspecs.org/#should are applied.

user_coffee_shop_query.rb
drqCode: Ahh sorry, I was referring to whole class (whole algorithm). I wanted to mark the entire file

beniaminmuresan: Please take a look at my latest updates. It is pretty obvious that sorting the distances array  would be bad if the csv file contains thousands of coffee shops. But I guess I was striving for readability and elegance of the code

drqCode: :1st_place_medal:

levelboy:
Followed up on email. Closing this

